PR #1, opened 2026-01-22
Add listAccounts, listFolders, and getRecentMessages tools

judicandus:
- Add listAccounts tool to list all configured email accounts
- Add listFolders tool to browse folder hierarchy with message counts
- Add getRecentMessages tool for efficient date-sorted message retrieval
- Enhance MCP bridge to handle resources/list and prompts/list
- Add comprehensive JSDoc documentation to all new functions
- Include CHANGELOG documenting all improvements

These additions provide structured access to Thunderbird accounts and folders, with fast recent message retrieval replacing slow full-mailbox searches. Messages are properly sorted by date and support filtering by folder, date range, and read status.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,0 +1,59 @@
+# Changelog
+
+## [Unreleased] - Enhanced Account and Folder Management
+
+### Added
+- **listAccounts** tool - Lists all configured email accounts with detailed information
+  - Returns account key, name, type (IMAP/POP3/RSS), email address, username, and hostname
+  - Useful for discovering available accounts before querying folders or messages
+  
+- **listFolders** tool - Lists all folders/mailboxes with filtering options
+  - Recursively traverses folder hierarchy across all accounts or specific account
+  - Returns folder name, path (URI), type (inbox/sent/drafts/trash/templates/folder)
+  - Includes message counts (total and unread) for each folder
+  - Properly detects folder types using Thunderbird's folder flags
+  
+- **getRecentMessages** tool - Efficiently retrieve recent messages with advanced filtering
+  - Date-based filtering (default: last 30 days, configurable)
+  - Returns messages sorted by date (newest first)
+  - Optional folder-specific search or automatic inbox-wide search
+  - Unread-only filter option
+  - Configurable result limit (default: 20, max: 100 messages)
+  - Automatically refreshes IMAP folders before reading
+  - Uses proper MIME decoding for international characters
+
+### Improved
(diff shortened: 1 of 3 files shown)


TKasperczyk (commented):
Thanks for the PR. listFolders, getRecentMessages, and the bridge fixes are useful additions. Some things need addressing before I can merge though:

### Folder flag values are wrong

In `listFolders`, the virtual folder check uses `0x1000`:

```js
if (folder.flags & 0x1000) return; // Ci.nsMsgFolderFlags.Virtual
```

`0x1000` is actually `nsMsgFolderFlags.Inbox`, not Virtual. Virtual is `0x00000020`. This means every Inbox folder gets silently skipped from results, which also makes the `"inbox"` type branch in the detection below it unreachable (same value, already filtered out).

Similarly, `0x00000800` is Queue (outbox), not Templates. The actual Templates flag is `0x00400000`.

Ref: https://searchfox.org/comm-central/source/mailnews/base/public/nsMsgFolderFlags.idl

### getRecentMessages vs extending searchMessages

`getRecentMessages` duplicates most of `searchMessages` (folder walking, IMAP refresh, mime2Decoded properties, same result shape). Date filtering and sort-by-date could be added as parameters to `searchMessages` instead of creating a parallel traversal function. Any fix to one would need to be mirrored in the other going forward.

The CHANGELOG's claim that it's "significantly faster" is also not accurate since both scan via `db.enumerateMessages()`. The only difference is the filter predicate.

### Bridge fixes are good

The `resources/list` and `prompts/list` handlers are correct and solve a real compatibility issue with clients like Claude Desktop. One thought: you could add the handlers *without* advertising the capabilities in `initialize`. Well-behaved clients won't call methods the server doesn't advertise, but misbehaving ones still get a clean response. Minor point though.

### Style

The codebase is intentionally minimal on comments - short inline notes for quirks/gotchas, no JSDoc blocks. The verbose `@param`/`@returns` documentation on every function doesn't match the rest of the code. The tool definitions at the top already serve as the API docs, so I'd ask to trim these down.

The CHANGELOG.md is also something I'd rather manage myself as the project evolves, so please drop it from the PR.

### Summary

The flag bugs are the blocker here. Once those are corrected (ideally tested against a real Thunderbird instance), and the style is aligned, we can discuss whether getRecentMessages should be a standalone tool or parameters on searchMessages. The bridge fix is ready to go and could be split into its own small PR if you want a quick merge on that part.

TKasperczyk:
@rwrz to answer your question: check out my comment in https://github.com/TKasperczyk/thunderbird-mcp/pull/13

PR #10, opened 2026-02-12
fix: handle HTML-only emails and Unicode encoding in JSON responses

quittung:
## Summary

Two related fixes for message reading reliability:

- **HTML body fallback**: `coerceBodyToPlaintext()` throws for some HTML-only emails (e.g. transactional emails from PayPal, banks, etc.). When it fails or returns empty, the fix walks the MIME tree using `part.contentType` to find `text/plain` or `text/html` parts. If only HTML is found, it strips tags, common entities and whitespace to produce readable plain text.

- **JSON encoding**: Thunderbird's `httpd` writes response strings as raw bytes (each JS char code = one byte). Unicode characters above U+007F whose low byte equals `0x22` (`"`) would corrupt the JSON output — e.g. the bullet character `•` (U+2022) has low byte `0x22`, which inserts a literal `"` into the byte stream and breaks JSON parsing. Wrapping both `res.write()` calls with `sanitizeForJson()` ensures the entire JSON response is safe, not just individual field values.

## Test plan

- [ ] Open an HTML-only email (e.g. a PayPal payment confirmation) and call `getMessage` — body should contain readable text instead of `"(Could not extract body text)"`
- [ ] Search for messages from senders whose names/subjects contain non-ASCII characters (umlauts, accented chars, bullet `•`, smart quotes, etc.) — the response should be valid JSON
- [ ] Confirm `searchMessages` with broad/empty queries no longer returns `"Tool execution failed"` due to JSON parse errors in the bridge

🤖 Generated with [Claude Code](https://claude.ai/claude-code)

### DIFF
--- a/extension/mcp_server/api.js
+++ b/extension/mcp_server/api.js
@@ -534,8 +534,44 @@ var mcpServer = class extends ExtensionCommon.ExtensionAPI {
                       // sanitizeForJson removes control chars that break JSON
                       body = sanitizeForJson(aMimeMsg.coerceBodyToPlaintext());
                     } catch {
-                      body = "(Could not extract body text)";
+                      // coerceBodyToPlaintext fails for some HTML-only messages
                     }
+                    if (!body) {
+                      function findTextPart(part) {
+                        const ct = (part.contentType || part.type || "").split(";")[0].trim().toLowerCase();
+                        if (ct === "text/plain" && part.body) return { content: part.body, isHtml: false };
+                        if (ct === "text/html" && part.body) return { content: part.body, isHtml: true };
+                        if (part.parts) {
+                          for (const sub of part.parts) {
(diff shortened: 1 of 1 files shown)


gui-wf:
Thanks for this. I maintain a downstream fork ([gui-wf/thunderbird-api](https://github.com/gui-wf/thunderbird-api)) and integrated both fixes from this PR:

1. Wrapping `res.write()` in `sanitizeForJson()` to prevent the low-byte `0x22` collision - this was a real bug I hadn't caught.
2. The `findTextPart` MIME tree walker with HTML tag stripping for HTML-only emails.

I also added a `bodyIsHtml` metadata flag to the `getMessage` response (inspired by PR #12) so callers can tell which extraction path was used.

PR #12, opened 2026-02-17
fix(getMessage): fall back to HTML body when plain text extraction fails

lihaokun:
## Summary
- When `coerceBodyToPlaintext()` throws (e.g. for forwarded Gmail messages with no plain text MIME part), `getMessage` now recursively walks the MIME parts to find a `text/html` or `text/plain` body instead of returning `"(Could not extract body text)"`.
- Adds a `bodyIsHtml` boolean field to the response so callers know whether to parse HTML.

## Motivation
Forwarded emails (especially from Gmail) often have only an HTML body embedded in a `multipart/mixed` or nested `message/rfc822` structure. `coerceBodyToPlaintext()` cannot handle these and throws, making the email body inaccessible via MCP.

## Test plan
- [x] Tested with forwarded Gmail emails that previously returned "(Could not extract body text)"
- [x] Verified normal emails still return plain text with `bodyIsHtml: false`
- [x] Verified HTML body is properly sanitized via `sanitizeForJson()`

### DIFF
--- a/extension/mcp_server/api.js
+++ b/extension/mcp_server/api.js
@@ -148,6 +148,30 @@ var mcpServer = class extends ExtensionCommon.ExtensionAPI {
           required: ["messageId", "folderPath", "to"],
         },
       },
+      {
+        name: "markAsRead",
+        title: "Mark As Read",
+        description: "Mark one or more messages as read (or unread). Accepts a single message or an array of messages.",
+        inputSchema: {
+          type: "object",
+          properties: {
+            messages: {
+              type: "array",
+              items: {
+                type: "object",
+                properties: {
+                  messageId: { type: "string", description: "The message ID (from searchMessages results)" },
+                  folderPath: { type: "string", description: "The folder URI path (from searchMessages results)" },
+                },
+                required: ["messageId", "folderPath"],
+              },
+              description: "Array of {messageId, folderPath} objects to mark",
+            },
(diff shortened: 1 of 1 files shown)


TKasperczyk:
Merged, thanks for this. The ccList addition, field-level sanitizeForJson, mime2Decoded properties in getMessage, and bodyIsHtml flag were all good improvements.

Fixed a few things on top in c4308dd:

- **findBody preference order**: it was checking text/html before text/plain, so in a multipart/alternative where HTML came first in the tree, it'd grab that regardless. Flipped the priority and changed the sibling scan to return plain immediately but track HTML as a fallback.
- **HTML stripping**: the raw HTML body wasn't super useful for LLM consumers, so I ported the tag-stripping approach from PR #10. Removes script/style blocks, converts block tags to newlines, single-pass entity decoder (handles hex entities too). Since we strip the HTML, bodyIsHtml is now always false when that path runs.
- **findBody contentType matching**: was doing exact `=== "text/html"` which breaks if the MIME type includes parameters like `; charset=utf-8`. Now strips params before comparing.
- **markAsRead performance**: was doing O(n) enumerateMessages per message. Now groups by folder (one db open each), tries getMsgHdrForMessageID first, falls back to building a hdr map once per folder. Also added the IMAP updateFolder refresh pattern from searchMessages.
- **markAsRead input handling**: accepts a single {messageId, folderPath} object now (normalizes to array), validates each item's shape, always returns an array.

PR #15, opened 2026-02-28
fix: proper UTF-8 encoding via TextEncoder, remove sanitizeForJson

lihaokun:
## Summary

- **Root cause fix**: `httpd.sys.mjs`'s `write()` method uses `bodyOutputStream.write()` which treats JavaScript strings as Latin-1, only preserving the low byte of each UTF-16 code unit. This silently corrupts any non-Latin-1 text (CJK characters, accented characters, etc.)
- **Fix**: Use `TextEncoder` to properly encode strings as UTF-8 before writing to the output stream
- **Cleanup**: Remove `sanitizeForJson()` function and all ~20 call sites — no longer needed since `JSON.stringify` already escapes control characters, and `TextEncoder` handles encoding correctly

## Changes

| File | Change |
|------|--------|
| `extension/httpd.sys.mjs` | `TextEncoder`-based `write()` for correct UTF-8 encoding |
| `extension/mcp_server/api.js` | Remove `sanitizeForJson` function and all call sites (-77 lines) |
| `scripts/build.sh` | Clean old XPI before rebuild to prevent stale artifacts |

## Testing

Verified all tools return correct CJK text after changes:
- `listFolders` — Chinese folder names (垃圾规则起效, 身份证, 会议, 夏老师, etc.)
- `searchMessages` — Chinese subjects and authors
- `getRecentMessages` — Chinese subjects and folder paths
- `getMessage` — Chinese email body content
- `listAccounts` — account names

## Context

`sanitizeForJson` was introduced in PR #12 as a workaround for the Latin-1 encoding issue. This PR fixes the root cause, making the workaround unnecessary.

### DIFF
--- a/extension/httpd.sys.mjs
+++ b/extension/httpd.sys.mjs
@@ -3818,7 +3818,18 @@ Response.prototype = {
     }
 
     var dataAsString = String(data);
-    this.bodyOutputStream.write(dataAsString, dataAsString.length);
+    // Encode as UTF-8 to correctly handle non-ASCII characters.
+    // The default bodyOutputStream.write() only takes the low byte of each
+    // UTF-16 code unit, silently discarding the high byte and corrupting
+    // any non-Latin-1 text (e.g., CJK characters).
+    var encoder = new TextEncoder();
+    var utf8Bytes = encoder.encode(dataAsString);
+    var binaryChars = new Array(utf8Bytes.length);
+    for (var i = 0; i < utf8Bytes.length; i++) {
+      binaryChars[i] = String.fromCharCode(utf8Bytes[i]);
+    }
+    var binaryStr = binaryChars.join('');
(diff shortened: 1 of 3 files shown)


TKasperczyk:
Merged, thanks. Good root cause fix. The TextEncoder approach is the right way to handle this, fixing it at the write layer instead of working around it per-field.

Verified on a live instance with Polish folder names and emoji subjects/bodies (4-byte UTF-8). Everything comes through clean.

One minor note for context: the old sanitizeForJson also stripped control characters (0x00-0x08, 0x0B, 0x0C, 0x0E-0x1F, 0x7F). With it removed, those now pass through to JSON.stringify which escapes them as \uXXXX. That's arguably more correct since it preserves data instead of silently dropping it. 0x7F (DEL) isn't escaped by JSON.stringify at all, but that's harmless in practice.

PR #32, opened 2026-03-06
fix: prevent GC of nsITimer in composeMail, set attachment size

lihaokun:
## Problem

When calling \`sendMail\` / \`replyToMessage\` / \`forwardMessage\` with attachments, the attachments were silently not added to the compose window.

Root cause: \`nsITimer\` was stored in a local variable. After the function returned normally, the variable went out of scope and the timer was garbage collected before its callback fired.

This explained a puzzling observation: a version with a **subsequent JS error** (\`failedAttachments is not defined\`) actually *worked* — because the error kept the JS engine busy past the timer registration, preventing GC.

A secondary issue: attachment size was shown as 0 bytes because \`att.size\` was never set.

## Fix

- **\`_attachTimers\` Set**: replace the single \`_attachTimer\` ref with a module-level \`Set\` so concurrent compose operations each keep their own timer alive.
- **\`injectAttachmentsAsync(attachDescs)\`**: shared helper used by all three compose functions. Accepts \`{ url, name, size?, contentType? }\` descriptors to handle both local file paths and original message attachment URLs (needed for forward).
- **\`filePathsToAttachDescs(filePaths)\`**: validates file existence upfront and returns \`{ descs, failed }\`. Missing files are reported synchronously in the tool response.
- **Remove dead \`addAttachments(composeFields, ...)\`**: this function silently dropped attachments on \`nsIMsgCompType.New\` and is fully replaced.
- **Fix \`replyToMessage\` and \`forwardMessage\`**: both now use \`injectAttachmentsAsync\`. Forward passes original message attachments alongside user-specified files.
- **\`COMPOSE_WINDOW_LOAD_DELAY_MS\`**: named constant for the 1500ms delay.
- **\`createLocalFile(path)\`**: small helper to avoid repeating the \`nsIFile\` instantiation pattern.

## Testing

Verified all three functions with a real attachment:
- \`sendMail\` + attachment ✅
- \`replyToMessage\` + attachment ✅  
- \`forwardMessage\` + attachment ✅

Attachment appears in compose window with correct filename and size.

### DIFF
--- a/extension/mcp_server/api.js
+++ b/extension/mcp_server/api.js
@@ -21,6 +21,8 @@ const resProto = Cc[
 const MCP_PORT = 8765;
 // Keep references to active attach timers to prevent GC before they fire.
 const _attachTimers = new Set();
+// Delay before injecting attachments into a newly opened compose window.
+const COMPOSE_WINDOW_LOAD_DELAY_MS = 1500;
 const DEFAULT_MAX_RESULTS = 50;
 const MAX_SEARCH_RESULTS_CAP = 200;
 const SEARCH_COLLECTION_CAP = 1000;
@@ -554,6 +556,13 @@ var mcpServer = class extends ExtensionCommon.ExtensionAPI {
               return null;
             }
 
+            /** Creates an nsIFile instance for the given path. */
+            function createLocalFile(path) {
+              const file = Cc["@mozilla.org/file/local;1"].createInstance(Ci.nsIFile);
+              file.initWithPath(path);
+              return file;
+            }
+
             /**
              * Converts local file paths to attachment descriptors, validating existence upfront.
              * Returns { descs: [{url, name, size}], failed: string[] }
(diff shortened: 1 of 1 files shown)


lihaokun:
Thanks for the detailed review! All four points addressed:

1. **Set of timers**: replaced `_attachTimer` with `_attachTimers = new Set()` so concurrent operations each hold their own reference.

2. **Shared helper for all three functions**: extracted `injectAttachmentsAsync(attachDescs)` which accepts `{ url, name, size?, contentType? }` descriptors — this lets `forwardMessage` pass original message attachment URLs alongside user-specified local files through the same path.

3. **Error reporting**: added `filePathsToAttachDescs(filePaths)` which validates file existence upfront and returns `{ descs, failed }`. Missing files are now reported synchronously in the tool response message. (Injection failures after the window opens are inherently async and can't be surfaced — but if the file exists at call time, injection should always succeed.)

4. **Dead code removed**: `addAttachments(composeFields, ...)` is gone. Also extracted `createLocalFile(path)` to avoid repeating the `nsIFile` instantiation pattern (it was duplicated in `ensureAttachmentDir` too), and named the `1500` delay as `COMPOSE_WINDOW_LOAD_DELAY_MS`.

Tested `sendMail`, `replyToMessage`, and `forwardMessage` all with attachments — all working.

PR #59, opened 2026-03-25
security: fix XSS, DoS, symlink, and permission issues

andrewboldi:
## Summary

Fixes 11 security issues found during a comprehensive audit (including hidden unicode scan).

### Commit 1: Core fixes
- **XSS in options.js** (Medium): Error messages were concatenated into `innerHTML`, allowing potential script injection. Replaced with `textContent` via DOM API.
- **Pre-auth memory exhaustion DoS** (Medium): No request body size limit — any local process could send a huge POST to exhaust Thunderbird's memory. Added 10 MB limit in both `httpd.sys.mjs` (pre-buffer rejection via HTTP 413) and the `api.js` handler (defense-in-depth).
- **Symlink attack on tmp directory** (Medium): `writeConnectionInfo` didn't check if `/tmp/thunderbird-mcp` was a symlink before writing. Added `isSymlink()` check.
- **Attachment dirs world-readable** (Low): `ensureAttachmentDir` created directories with `0o755` while other temp dirs used `0o700`. Fixed to `0o700`.
- **Incomplete `escapeHtml`** (Low): Missing `"` and `'` escaping. Added `&quot;` and `&#39;`.
- **Connection file port not validated** (Low): Bridge accepted any truthy value for port. Added integer range validation (1-65535).
- **`sanitizeJson` double-backslash handling** (Low): Lookbehind didn't handle double-escaped backslashes correctly. Replaced with capturing-group approach.

### Commit 2: Additional findings from deep review
- **Unauthenticated `/trace` endpoint** (Medium): `httpd.sys.mjs` had a default `/trace` handler that echoed all request headers — including the `Authorization` bearer token — without any authentication. Removed entirely.
- **Auth bypass on non-POST requests** (Low): Non-POST requests returned 200 OK without checking auth, allowing unauthenticated server probing. Moved auth check before method check; non-POST now returns 405 with `Allow: POST` header.
- **`web_accessible_resources` exposure** (Low): `httpd.sys.mjs` was listed in `web_accessible_resources` but is loaded via `resource://` protocol. Removed to prevent extension fingerprinting.
- **Saved attachment files world-readable** (Low): `file.createUnique` used `0o644`. Fixed to `0o600`.

### Also confirmed clean
- **No hidden unicode characters** — exhaustive scan for bidi overrides, zero-width chars, homoglyphs, Cyrillic, fullwidth, tag characters found nothing malicious.

## Test plan

- [x] All 275 existing tests pass
- [ ] Verify options page error rendering still works with DOM API approach
- [ ] Test oversized POST (>10 MB) returns 413 instead of consuming memory
- [ ] Verify connection file writing fails if tmp dir is a symlink
- [ ] Verify saved attachments directory and files have 0o700/0o600 permissions
- [ ] Verify GET request to server returns 403 (not 200) without auth token
- [ ] Verify /trace endpoint returns 404

### DIFF
--- a/extension/httpd.sys.mjs
+++ b/extension/httpd.sys.mjs
@@ -1498,6 +1498,12 @@ RequestReader.prototype = {
           : 0;
         dumpn("_processHeaders, Content-length=" + this._contentLength);
 
+        // Reject oversized request bodies before buffering into memory.
+        // 10 MB is generous for JSON-RPC MCP requests.
+        if (this._contentLength > 10 * 1024 * 1024) {
+          throw HTTP_413;
+        }
+
         this._state = READER_IN_BODY;
       }
       return done;
--- a/extension/httpd.sys.mjs
+++ b/extension/httpd.sys.mjs
@@ -3441,6 +3441,9 @@ ServerHandler.prototype = {
   /**
    * Contains handlers for the default set of URIs contained in this server.
    */
+  // /trace endpoint removed — it echoed all request headers (including
+  // Authorization tokens) without authentication, enabling token leakage
+  // to any local process.
   _defaultPaths: {
     "/": function (metadata, response) {
       response.setStatusLine(metadata.httpVersion, 200, "OK");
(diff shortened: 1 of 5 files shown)


andrewboldi:
### Code review

No issues found. Checked for bugs and CLAUDE.md compliance.

🤖 Generated with [Claude Code](https://claude.ai/code)

TKasperczyk:
Thanks @andrewboldi -- merged in 2f661bc with two small follow-up fix commits on top.

All 11 of your security items are in:
- XSS in options.js (DOM API instead of innerHTML) ✓
- /trace endpoint removed (it was echoing Authorization headers, legitimate find) ✓
- Auth-before-method reorder ✓
- Symlink check on <TmpD>/thunderbird-mcp ✓
- Attachment dirs 0o700, files 0o600 ✓
- escapeHtml escaping " and ' ✓
- Port range validation in bridge ✓
- sanitizeJson regex rewrite (verified correct and no ReDoS risk) ✓
- web_accessible_resources cleared ✓
- Pre-auth body size cap ✓

Post-merge review surfaced two items I addressed in [fb9d38b](https://github.com/TKasperczyk/thunderbird-mcp/commit/fb9d38b):

1. **HTTP body cap raised from 10 MB to 32 MB.** The project already has a `MAX_BASE64_SIZE = 25 MB` contract for inline attachments in `sendMail`. Your 10 MB cap in both `httpd.sys.mjs` and `api.js` was silently breaking that -- any attachment larger than ~7 MB (after JSON-RPC framing) would have hit HTTP 413 before reaching the 25 MB base64 size check. Aligned both caps at 32 MB, which still caps the DoS surface at a round value while preserving the existing attachment functionality.

2. **Stale `/trace` reference in a doc comment.** Cosmetic: the `ServerHandler` module comment still described `/trace` as a default path even after you removed the handler. Updated.

A post-merge security review also flagged the symlink hardening as "partial" (you only checked `tmpDir` in `writeConnectionInfo`, not in the other `TmpD/thunderbird-mcp/...` code paths like `ensureAttachmentDir`). After consideration I decided not to extend it -- `writeConnectionInfo` runs at MCP server start and throws on symlink, so if the check fires the server never runs and the other paths don't execute. Narrow TOCTOU windows between your check and subsequent operations would be partial hardening anyway. Leaving as-is.

Solid work on the audit. Thanks for the /trace catch especially -- that one was a real vuln.

PR #60, opened 2026-03-25
fix: detect Snap-sandboxed Thunderbird tmp directory

andrewboldi:
## Summary
- When Thunderbird is installed via Snap, the sandbox redirects `TMPDIR` to a user-writable location (e.g. `~/Downloads/thunderbird.tmp`). The extension writes `connection.json` there via `Services.dirsvc.get("TmpD")`, but the bridge runs outside the sandbox so `os.tmpdir()` returns `/tmp` — causing a silent connection failure.
- Adds `detectSnapTmpDir()` that checks for a Snap install (`~/snap/thunderbird`), reads the running process's `TMPDIR` from `/proc/<pid>/environ`, and falls back to well-known Snap tmp paths.
- The standard `/tmp` path is tried first, so native installs are completely unaffected.

## Test plan
- [ ] Verify bridge connects successfully with Snap-installed Thunderbird
- [ ] Verify bridge still works with native (non-Snap) Thunderbird installs
- [ ] Verify fallback path works when `/proc` is not readable

### DIFF
--- a/mcp-bridge.cjs
+++ b/mcp-bridge.cjs
@@ -19,10 +19,69 @@ const CONNECTION_RETRY_DELAY_MS = 1000;
 const CONNECTION_MAX_RETRIES = 5;
 
 /**
+ * Detect the Thunderbird Snap tmp directory.
+ *
+ * When Thunderbird is installed via Snap, the sandbox redirects TMPDIR to a
+ * user-writable location (e.g. ~/Downloads/thunderbird.tmp). The extension
+ * writes its connection file there, but this bridge runs outside the sandbox
+ * so os.tmpdir() returns the system /tmp instead.
+ *
+ * We detect the Snap install by looking for ~/snap/thunderbird and then check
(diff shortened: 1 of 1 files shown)


TKasperczyk:
Thanks @andrewboldi -- closing this in favor of a broader sandbox synthesis merged as 2a025af.

Your `/proc/<pid>/environ` approach for reading Thunderbird's live TMPDIR is the right move for Snap and it's preserved in the new code (see `findSnapConnectionCandidates` in mcp-bridge.cjs). I combined it with Flatpak runtime-dir scanning (incl. Betterbird and Beta), a `THUNDERBIRD_MCP_CONNECTION_FILE` env-var override, a macOS `/var/folders` scan, and dynamic re-resolution on every cache miss (the previous one-shot resolution had a startup race where the bridge could lock onto a wrong path before Thunderbird finished creating the connection file).

Also added regression tests that cover Snap detection from a mocked `/proc` tree, the startup-race case, and a few others.

Appreciate the contribution -- the Snap detection logic was the strongest piece of the three sandbox PRs.

Issue #54, opened 2026-03-19
Can't see html emails


gasteigerjo:
Currently, the MCP server doesn't allow the model to see html-based emails that don't have a text version. This restricts functionality quite a bit.

I'd suggest to add this by converting html emails to markdown by default. Something like a Javascript version of [markdownify](https://github.com/matthewwithanm/python-markdownify) could be useful for that. It might be good to provide a parameter for the model to see the full html, in case that's needed for something. And an email response tool where the html is prepended but the model only sees a clear placeholder would also be useful.

btw, for my own MCP servers I found it useful to tell Claude to compare the functionality to similar MCP servers and then implement all the feature gaps. You could do that e.g. with [Outlook MCP](https://github.com/ryaker/outlook-mcp).

TKasperczyk:
Hi -- status update. PR #64 (merged in commit 3a7046d) adds a fallback path for messages where Thunderbird's structured MIME extraction returns empty results -- which is typically the case for imported mbox archives.

**What now works:**
- Singlepart `text/plain` mbox messages with any standard Content-Transfer-Encoding (7bit, 8bit, binary, quoted-printable, base64)
- Singlepart `text/html` mbox messages -- respecting the `bodyFormat` parameter (html/markdown/text) and extracting via the existing `stripHtml` / `htmlToMarkdown` helpers
- Arbitrary charsets -- decoded via TextDecoder with UTF-8 fallback for unknown labels
- Bare-CR line endings (classic pre-OSX Mac exports)

**What still doesn't work:**
- Multipart HTML emails from imported mbox archives. The fallback rejects `multipart/*` cleanly (logs and returns empty body) rather than trying to regex-parse MIME trees, which is what the original PR did and what made it unsafe to merge as-is. Full multipart support requires a real MIME parser and would be a significant undertaking.

If your "Can't see html emails" issue was about **singlepart** HTML mbox messages, this should fix it -- please pull latest main, rebuild the XPI, and confirm.

If it's about **multipart** HTML messages (one text/plain + one text/html alternative, or HTML with inline images), you'll still see empty output from this path. In that case please let me know the message structure (Thunderbird > View > Message Source will show it) and I can track that as a separate issue for proper multipart handling.

TKasperczyk:
Reopening -- this was auto-closed by GitHub's keyword parser picking up a 'close #54' substring in the #64 merge commit ('does not fully close #54'), which wasn't meant as an auto-close directive.

[...]

PR #64, opened 2026-03-30
Fix body extraction for local mbox folder messages

abedegno:
getMessage returns empty body for messages in local folders (imported mbox archives). Adds stream.available() fallback for zero messageSize, and raw stream body extraction fallback when MsgHdrToMimeMessage returns empty MIME parts. Tested with Outlook/Exchange mbox archives containing HTML-only emails.

### DIFF
--- a/extension/mcp_server/api.js
+++ b/extension/mcp_server/api.js
@@ -2817,9 +2817,19 @@ var mcpServer = class extends ExtensionCommon.ExtensionAPI {
 	                    try {
 	                      const folder = msgHdr.folder;
 	                      stream = folder.getMsgInputStream(msgHdr, {});
-	                      const messageSize = folder.hasMsgOffline(msgHdr.messageKey)
+	                      let messageSize = folder.hasMsgOffline(msgHdr.messageKey)
 	                        ? msgHdr.offlineMessageSize
 	                        : msgHdr.messageSize;
+	                      // For local folders (mbox), messageSize can be 0 or
+	                      // inaccurate for imported messages. Fall back to reading
+	                      // whatever is available in the stream.
+	                      if (!messageSize || messageSize <= 0) {
+	                        messageSize = stream.available();
+	                      }
+	                      if (!messageSize || messageSize <= 0) {
+	                        resolve({ error: "Message has zero size - cannot read raw source" });
+	                        return;
+	                      }
 	                      // No charset specified -- defaults to Latin-1 which
 	                      // preserves raw bytes. UTF-8 would corrupt messages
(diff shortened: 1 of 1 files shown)


TKasperczyk:
Thanks @abedegno -- merged in 3a7046d.

Your diagnosis of the root cause was right -- `MsgHdrToMimeMessage` returns empty parts for some imported mbox messages, and `messageSize` can be 0 for those. Both fallbacks are in.

The **rawSource size fallback** (lines that fall back to `stream.available()` when `messageSize` is 0) landed unchanged from your PR. Clean fix, good catch.

For the **body extraction fallback**, post-merge review flagged two serious issues with the original approach that needed rethinking:

1. **Whole-source substring match triggered wrong decoding for multipart.** The checks like `rawContent.toLowerCase().includes("content-transfer-encoding: base64")` matched ANY base64 header in the entire message, including attachment headers in multipart mail. That then triggered `atob()` on the whole body (including boundaries and attachment payloads), producing either exceptions or garbage binary output.

2. **`bodyFormat` contract broken.** Setting `bodyIsHtml = false` after HTML stripping meant `getMessage({ bodyFormat: "html" })` on an mbox HTML email would return plaintext, silently overriding what the caller asked for.

I rewrote the body-extraction fallback in [5ca78cb](https://github.com/TKasperczyk/thunderbird-mcp/commit/5ca78cb) + [0c9bb5f](https://github.com/TKasperczyk/thunderbird-mcp/commit/0c9bb5f) to:

- Parse **top-level headers only** (the header block before the first blank line), not substring scans on the whole message
- **Reject multipart** cleanly rather than try to regex-parse MIME trees
- Extract **charset** from Content-Type and decode via `TextDecoder` with a utf-8 retry for unknown charsets
- Handle 7bit / 8bit / binary / quoted-printable / base64 transfer encodings; fail closed on anything else
- **Preserve `bodyFormat`** -- html keeps raw HTML, markdown goes through the existing `htmlToMarkdown`, text/other through the existing `stripHtml`
- Close the stream in `finally`
- Handle bare-CR line endings (classic pre-OSX Mac mbox exports) in addition to CRLF and LF

This partially closes #54 -- singlepart mbox messages (text/plain or text/html) now extract correctly. Multipart HTML mbox messages still return empty body from this fallback because full multipart support needs a real MIME parser; the primary `MsgHdrToMimeMessage` path handles multipart when Thunderbird can parse it. Happy to discuss if you want to tackle the multipart case separately.

Thanks for driving this -- the zero-messageSize issue and the mbox-empty-MIME-tree issue were both real diagnostic wins.

PR #66, opened 2026-04-07
fix: update extension ID in install script

simon-77:
`install.sh` still used the old extension ID (`thunderbird-mcp@luthriel.dev`) after it was changed to `thunderbird-mcp@tkasperczyk.dev` in 8d2d397, causing silent installation failure — the XPI is copied with the wrong filename so Thunderbird doesn't recognize it.


TKasperczyk:
Thanks for catching this -- silent install failure after the extension ID rename is a nasty regression. Merging.

PR #73, opened 2026-04-16
feat(calendar): add updateTask tool

andrei-rizoiu:
## Summary

Adds a new `updateTask` MCP tool to update existing tasks/to-dos in Thunderbird calendars.

**Supported fields:**
- `taskId` *(required)* — task ID from `listTasks` results
- `calendarId` *(required)* — calendar ID from `listTasks` results
- `title` — new task title
- `dueDate` — new due date (ISO 8601)
- `description` — new task description/body
- `priority` — task priority
- `completed` — boolean; `true` sets `percentComplete=100` and records `completedDate`, `false` reopens the task
- `percentComplete` — numeric completion percentage

**Implementation notes:**
- Looks up the task by ID using `calendar.getItem()` with a fallback full-scan for compatibility across calendar backends
- Returns the updated task in the same format as `listTasks`
- Complements the existing `createTask` and `listTasks` tools to complete basic CRUD for calendar tasks

## Test plan

- [ ] Mark a task as completed (`completed: true`) and verify `percentComplete` becomes 100
- [ ] Reopen a completed task (`completed: false`) and verify it appears in `listTasks`
- [ ] Update `title` and `dueDate` and verify changes persist
- [ ] Confirm error is returned for unknown `taskId`

### DIFF
--- a/extension/mcp_server/api.js
+++ b/extension/mcp_server/api.js
@@ -270,6 +270,26 @@ var mcpServer = class extends ExtensionCommon.ExtensionAPI {
         },
       },
       {
+        name: "updateTask",
+        group: "calendar", crud: "update",
+        title: "Update Task",
+        description: "Update an existing task/to-do: change title, due date, description, priority, completion status, or percent complete",
+        inputSchema: {
+          type: "object",
+          properties: {
+            taskId: { type: "string", description: "Task ID (from listTasks results)" },
+            calendarId: { type: "string", description: "Calendar ID containing the task (from listTasks results)" },
+            title: { type: "string", description: "New task title (optional)" },
+            dueDate: { type: "string", description: "New due date in ISO 8601 format (optional)" },
+            description: { type: "string", description: "New task description/body (optional)" },
+            completed: { type: "boolean", description: "Set to true to mark the task done (sets percentComplete=100 and records completedDate), false to reopen it (optional)" },
+            percentComplete: { type: "integer", description: "Completion percentage 0–100 (optional)" },
+            priority: { type: "integer", description: "Priority: 1=high, 5=normal, 9=low (optional)" },
+          },
+          required: ["taskId", "calendarId"],
+        },
+      },
+      {
         name: "searchContacts",
         group: "contacts", crud: "read",
         title: "Search Contacts",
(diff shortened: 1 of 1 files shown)


TKasperczyk:
Thanks @andrei-rizoiu -- merged in 81fc411 with five follow-up fix commits on top of your work:

1. **`fix: support "integer" type in tool arg validator/coercer`** ([c31d255](https://github.com/TKasperczyk/thunderbird-mcp/commit/c31d255)) -- this one actually unblocked the tool. The existing validator did `typeof value !== expectedType`, but `typeof 50 === "number"` (not "integer"), so any call with `percentComplete: 50` or `priority: 5` would have been rejected before reaching `updateTask()`. Same latent bug in `listTasks.maxResults`; fix benefits both.

2. **`fix(updateTask): null dueDate clears field`** ([1eb9d68](https://github.com/TKasperczyk/thunderbird-mcp/commit/1eb9d68)) -- `new Date(null).getTime() === 0` which passed the `isNaN` check, so `dueDate: null` silently wrote 1970-01-01 instead of clearing. Now null/empty-string explicitly sets `newItem.dueDate = null`.

3. **`fix(updateTask): keep VTODO state consistent`** (same commit) -- setting `percentComplete` alone left `STATUS` stale (e.g. 100% complete but STATUS=IN-PROCESS), which Thunderbird's UI renders inconsistently. New `applyCompletionState()` helper updates all three iCal RFC 5545 fields (STATUS, PERCENT-COMPLETE, COMPLETED) together. Also rejects ambiguous calls that pass both `completed` and `percentComplete`.

4. **`fix(updateTask): check capabilities.tasks.supported`** ([eb74ee8](https://github.com/TKasperczyk/thunderbird-mcp/commit/eb74ee8)) -- matches the existing check in `createTask` and `listTasks`. Previously a calendar that doesn't support tasks would return a misleading "Task not found" error.

5. **`fix(updateTask): warn on recurring tasks`** (same commit) -- matches the warning `updateEvent` emits, so callers know the change hit the entire series rather than a single occurrence.

The core implementation -- `getItem` + full-scan fallback, `oldItem.clone()`, `modifyItem(newItem, oldItem)` -- is exactly right and matches Thunderbird's own task code. Nice work, and thanks for completing the task CRUD triad.

PR #80, opened 2026-04-24
fix: spec-compliant MCP transport + silent-failure sweep

Fuijin:
## Summary

Two related changes that surfaced from debugging an *Unable to connect to extension server* error in Claude Desktop ≥ 1.1.380:

1. **Bridge transport rewrite** so initialize responses actually reach modern Claude Desktop (`mcp-bridge.cjs`).
2. **Silent-failure sweep** across the extension and the bridge so the next mystery doesn't take an afternoon (`extension/mcp_server/api.js`, `extension/options.js`).

## Bridge changes (`mcp-bridge.cjs`)

- **Replace readline-based stdin loop with explicit utf-8 buffered `data` handler.** Under Claude Desktop's Electron-spawned Node on Windows the readline-based loop swallowed the initialize response — writes from promise callbacks never made it back through the pipe. The new loop matches what `@modelcontextprotocol/sdk` does and works reliably. Drops the private `process.stdout._handle.setBlocking()` call.
- **Spec-compliant initialize per MCP 2025-06-18 lifecycle.** Respond with the requested `protocolVersion` if known, else the latest version we support. `SUPPORTED_PROTOCOL_VERSIONS` keeps unknown future versions from being echoed blindly.
- **`serverInfo.version` from `package.json`** so a single bump propagates.
- **Gated stderr diagnostics** (`THUNDERBIRD_MCP_DEBUG=1`) for transport events. Spec allows it; clients are free to ignore.
- **`package.json` bumped to 0.5.0** to match the extension manifest.

## Extension changes (`extension/mcp_server/api.js`, `extension/options.js`)

Same spec-compliant initialize negotiation in the HTTP handler so callers that bypass the bridge get a current version. `serverInfo.version` resolves from `extension/manifest.json` lazily.

Silent-failure fixes (audited via a separate triage):

- `addAttachmentsToComposeWindow` no longer silently no-ops when the compose window is missing or `AddAttachments` is not a function — both branches log a warning. *Reports of \"email sent without attachments\" are now traceable.*
- `injectAttachmentsAsync` fire-and-forget timer logs caught exceptions instead of swallowing.
- Tool-dispatch IIFE in the HTTP handler has a top-level `.catch` so an unhandled rejection in the inner catch (or in `res.finish`) doesn't hang the connection.
- Both `server.stop` cleanup paths during `start()` failure now log instead of `catch {}`.
- `coerceToolArgs` array-parse failure warns with key + parse error.
- `getServerInfo` build-info / connection-info reads warn on unexpected errors (file-not-found stays quiet).
- `getAccountAccessConfig` warns on a corrupt `PREF_ALLOWED_ACCOUNTS` pref.
- `deleteAllMessagesRecursive` logs each per-folder failure with the folder URI.
- `options.js` page-load kickoffs (`loadServerInfo`, `loadAccountAccess`, `loadToolAccess`, `loadSkipReviewPref`) chain `.catch` safety nets.

## Repro of the original 60s timeout (before the bridge rewrite)

```bash
printf '%s\n' '{\"method\":\"initialize\",\"params\":{\"protocolVersion\":\"2025-11-25\",\"capabilities\":{},\"clientInfo\":{\"name\":\"x\",\"version\":\"1\"}},\"jsonrpc\":\"2.0\",\"id\":0}' | node mcp-bridge.cjs
```

Claude Desktop log (`%APPDATA%\Claude\logs\mcp-server-Thunderbird MCP.log`) showed:

```
... Message from client: {\"method\":\"initialize\",\"params\":{\"protocolVersion\":\"2025-11-25\",...}}
... Message from client: {\"method\":\"notifications/cancelled\",\"params\":{\"reason\":\"Request timed out\"}}
```

Filesystem MCP server in the same Claude Desktop install responds with `\"protocolVersion\":\"2025-06-18\"` (an older version) and Claude attaches normally — confirming version mismatch was a red herring; the real fault was that the bridge's initialize response never reached the parent process under the readline loop.

## Test plan

- [x] Existing test suite passes (263/282; 1 pre-existing Windows uid-test failure, 18 skipped because TB is running)
- [x] `node --check` clean on bridge, api.js, options.js
- [x] Local stdio repro of all bridge-handled methods (initialize known + unknown version, ping, resources/list, prompts/list, notifications/initialized, tools/list)
- [ ] Reviewer to load updated XPI in TB and confirm Claude Desktop attaches without timeout

## Notes for follow-up

A modular split of `api.js` (currently 5856 LoC) into `extension/mcp_server/lib/*.sys.mjs` is in progress on a separate branch and will land as its own PR sequence — not folded into this one to keep review tractable.

### DIFF
--- a/extension/mcp_server/api.js
+++ b/extension/mcp_server/api.js
@@ -4628,7 +4628,10 @@ var mcpServer = class extends ExtensionCommon.ExtensionAPI {
                     count += hdrs.length;
                   }
                 }
-              } catch {}
+              } catch (e) {
+                // Continue traversal; log per-folder failures so partial empties are visible.
+                console.error("thunderbird-mcp: deleteMessages failed for folder", folder?.URI || folder?.name, ":", e);
+              }
               if (folder.hasSubFolders) {
                 for (const sub of folder.subFolders) {
                   count += deleteAllMessagesRecursive(sub);
@@ -5305,7 +5308,10 @@ var mcpServer = class extends ExtensionCommon.ExtensionAPI {
                   try {
                     const parsed = JSON.parse(value);
                     if (Array.isArray(parsed)) args[key] = parsed;
-                  } catch { /* leave as-is for validation to catch */ }
+                  } catch (e) {
+                    // Leave value as-is so validator surfaces a typed error to the client.
+                    console.warn(`thunderbird-mcp: coerceToolArgs JSON.parse failed for key=${key}:`, e.message);
+                  }
                 }
               }
               return args;
@@ -5559,7 +5565,10 @@ var mcpServer = class extends ExtensionCommon.ExtensionAPI {
                   }));
                 }
                 res.finish();
-              })();
+              })().catch((e) => {
+                console.error("thunderbird-mcp: unhandled dispatch error:", e);
+                try { res.finish(); } catch {}
+              });
             });
 
             // Try the default port first, then fall back to nearby ports
@@ -5584,7 +5593,7 @@ var mcpServer = class extends ExtensionCommon.ExtensionAPI {
               connFilePath = writeConnectionInfo(boundPort, authToken);
             } catch (writeErr) {
               // Connection file write failed -- stop the orphaned server
-              try { server.stop(() => {}); } catch {}
+              try { server.stop(() => {}); } catch (e) { console.error("thunderbird-mcp: server.stop failed:", e); }
               globalThis.__tbMcpServer = null;
               throw writeErr;
             }
@@ -5595,7 +5604,7 @@ var mcpServer = class extends ExtensionCommon.ExtensionAPI {
             console.error("Failed to start MCP server:", e);
             // Stop server if it was started but something else failed
             if (globalThis.__tbMcpServer) {
-              try { globalThis.__tbMcpServer.stop(() => {}); } catch {}
+              try { globalThis.__tbMcpServer.stop(() => {}); } catch (e) { console.error("thunderbird-mcp: server.stop failed:", e); }
               globalThis.__tbMcpServer = null;
             }
             // Clear cached promise so a retry can attempt to bind again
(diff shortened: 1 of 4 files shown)


TKasperczyk:
Merged, thanks. The bridge transport rewrite was the main win -- the readline-vs-buffered-data diagnosis is solid and the repro made the failure mode reproducible in seconds. Silent-failure sweep was a useful bonus.

Followed up in d25d1d3 with four small fixes from a code review:

- **Legacy protocol version**: added `2024-10-07` to `SUPPORTED_PROTOCOL_VERSIONS` in both the bridge and the HTTP handler. `@modelcontextprotocol/sdk` v1.x still advertises that version, so clients on that SDK line would have gotten `2025-11-25` echoed back and disconnected.
- **Strict protocolVersion validation**: missing/non-string `protocolVersion` now returns `-32602 invalid params` instead of falling through to LATEST. Matches the MCP 2025-06-18 schema and the TS SDK's pre-negotiation validation.
- **Attachment-conversion logging**: added a `console.warn` in `descsToMsgAttachments` (api.js:1389). The sweep covered the compose-window helpers but missed this -- it's also the underlying source of "email sent without attachments" on the direct-send path, since `sendMessageDirectly` calls into it too.
- **Version bump**: 0.5.0 -> 0.5.1 in `package.json` and `extension/manifest.json`. v0.5.0 had already been tagged/released before this PR opened.

Not addressed: the HTTP endpoint isn't true Streamable HTTP per 2025-06-18 (no `MCP-Protocol-Version` request-header handling, no HTTP 400 on bad versions). Effectively bridge-private though, so I'd rather narrow the spec-compliance framing than implement that surface for callers we don't have.

Fuijin:
I have a draft in a different branch. But, this thing is a bigger accessebility rewrite and more loose coupling thing as just had some fun with the whole project for my company platform as I needed some Bulk imports and batching an dispatching via Halon or Postfix.

PR #81, opened 2026-04-24
refactor(api): split api.js monolith into lib/*.sys.mjs (5 phases, 5856 -> 1130 LoC, -81%)

Fuijin:
## Summary

Splits the monolithic 5856-line `extension/mcp_server/api.js` into 11 focused modules under `extension/mcp_server/lib/`. The remaining api.js (~1130 LoC) holds only the WebExtension experiment-API class shell, lazy SDK imports, the factory wiring, the mcpServer namespace methods, the inline HTTP request handler closure (it captures `res` / `server` and isn't worth threading through a factory), and `onShutdown`.

5 phases, each independently shippable. Final layout:

```
extension/mcp_server/lib/
├── constants.sys.mjs       (60 LoC)   protocol versions, size limits,
│                                       pref keys, group/CRUD validators,
│                                       tag blacklist
├── state.sys.mjs           (30 LoC)   _attachTimers, _tempAttachFiles,
│                                       _claimedReplyComposeWindows
├── tools-schema.sys.mjs   (685 LoC)   the 36-tool inputSchema array
├── auth.sys.mjs           (~100 LoC)  generateAuthToken,
│                                       writeConnectionInfo,
│                                       removeConnectionInfo,
│                                       timingSafeEqual
├── access-control.sys.mjs (~150 LoC)  account/folder/tool/skipReview
│                                       policy backed by TB prefs
├── accounts.sys.mjs        (80 LoC)   listAccounts, findIdentity*,
│                                       getIdentityAutoRecipientHeader
├── contacts.sys.mjs       (187 LoC)   address-book CRUD
├── folders.sys.mjs        (498 LoC)   folder traversal + CRUD;
│                                       openFolder = getAccessibleFolder
│                                       + IMAP refresh + msgDatabase
├── calendar.sys.mjs       (723 LoC)   events + tasks; tolerates null
│                                       cal/CalEvent/CalTodo on TB builds
│                                       without the calendar component
├── filters.sys.mjs        (509 LoC)   mail rules; preserves the
│                                       MailServices.filters -> XPCOM
│                                       contract-id fallback
├── messages.sys.mjs      (1492 LoC)   search/get/list/update/delete with
│                                       Gloda + MIME parsing + attachment
│                                       offline-cache handling
├── compose.sys.mjs       (1003 LoC)   sendMail/replyToMessage/
│                                       forwardMessage with attachment
│                                       timers, compose-window observers,
│                                       sendMessageDirectly TB version
│                                       fallback (16-arg vs 18-arg)
└── dispatch.sys.mjs       (238 LoC)   validateToolArgs / coerceToolArgs
                                        / callTool flat-routing through
                                        a toolHandlers bag the caller
                                        assembles
```

## Pattern: `make<Module>({ deps })`

`Services` / `Cc` / `Ci` / `MailServices` are sandbox globals in api.js (the WebExtension experiment-API context) but not in `.sys.mjs` files. Plain top-level imports of these globals would fail. Each module exports a `make<Module>({ deps })` factory; api.js calls each factory once with the dependencies it needs and gets back an operational object. This decouples each module file from however Mozilla exposes those globals across TB versions and makes the dependency graph explicit.

Domain layering (also enforced as a comment in each module header):

```
0: state, constants, tools-schema      (no XPCOM)
1: auth, access-control                (XPCOM but no cross-domain deps)
2: accounts, folders, contacts,
   calendar, filters                   (uses access-control)
3: messages                            (uses folders + access-control)
4: compose                             (uses accounts + folders + messages
                                        + access-control)
5: dispatch                            (uses every domain via toolHandlers)
```

No cycles, verified by `scripts/check-imports.cjs`.

## Tooling shipped with this PR

`scripts/check-imports.cjs` -- AST-based static module-graph validator (acorn). Walks every `.js` / `.mjs` / `.sys.mjs` / `.cjs` under `extension/`, extracts both `import { ... } from "..."` and `ChromeUtils.importESModule(...)` calls including the destructured form, resolves `resource://thunderbird-mcp/...` URLs against on-disk files, asserts file existence and named-symbol presence, and detects cycles via DFS coloring. Useful as a pre-commit hook. Adds `acorn` as the only new devDependency (single small package, no transitive deps).

Output on the final tree: `Module graph OK: 17 files, 14 thunderbird-mcp imports, cycles=0.`

## Smoke test

Phases 1+2 were live-smoke-tested in real Thunderbird against the rebuilt XPI via Claude Desktop -- listAccounts (10 accounts), listCalendars (4 calendars), getAccountAccess, searchMessages countOnly returning the SEARCH_COLLECTION_CAP=10000 from constants.sys.mjs, listFilters, full searchMessages with Gloda body preview. All passed.

Phases 3-5 are NOT yet smoke-tested in a live Thunderbird (the running TB still holds the phase-1+2 XPI). I verified each new module with `node --check` and the AST validator, and every extracted function is byte-faithful from the original (acorn-driven extraction script identifies exact line ranges, copies the source text verbatim, wraps in the factory). But before merging please reload the XPI and re-run a smoke loop -- especially the compose path (sendMail / replyToMessage / forwardMessage with attachments, the most fragile area).

## Migration order rationale (and what to revert if something breaks)

Each phase is its own commit so any regression bisects cleanly:

1. `786b13a` Phase 1 -- pure data extraction (constants, state, tools-schema). Lowest risk.
2. `27c312e` AST-based import validator. Tooling, no behavior change.
3. `3d22519` Phase 2 -- auth + access-control. Leaf modules with light XPCOM deps.
4. `795c247` Phase 3+4 -- accounts, contacts, folders, calendar, filters, messages. Domain leaves + the big messages cluster.
5. `cccd4ae` Phase 5 -- compose + dispatch. Deepest closures + dead-code sweep of orphaned constants from earlier phases.

## Build

`scripts/build-xpi.cjs` does an unconditional recursive copy of `extension/`, so all new `lib/*.sys.mjs` files are picked up automatically -- no build-script change needed. XPI sizes: 215 KB (pre-refactor) -> 228 KB (post-refactor). The slight growth is the .sys.mjs files compressing less well than the inline bodies, expected.

## Dependency note

This PR is stacked on #80 (the bridge transport fix + silent-failure sweep). If #80 lands first, GitHub will collapse this PR's diff to just the modular changes.


Fuijin:
Live attach now confirmed working. Posting the runtime trace from a real Claude Desktop spawn here so the diagnosis is on the record (commit \`cce4d9b\`):

\`\`\`
LOAD pid=21428 ppid=20372 node=v24.14.0 platform=win32 cwd=C:\WINDOWS\system32
LOAD argv=["node.exe","...\Claude Extensions\local.mcpb.tomasz-kasperczyk.thunderbird-mcp/mcp-bridge.cjs"]
LOAD require.main===module: false
LOAD require.main path: C:\Program Files\WindowsApps\Claude_1.3883.0.0_x64__pzs8sxrjxfjjc\app\resources\app.asar\.vite\build\mcp-runtime\nodeHost.js
LOAD module.filename: ...\mcp-bridge.cjs
LOAD getuid available: false (Windows always false)
LOAD stdin.isTTY=false stdout.isTTY=false stderr.isTTY=false
AUTOSTART decision=true (require.main===module: false, argv[1]_basename=mcp-bridge.cjs)
STARTBRIDGE entered
STARTBRIDGE listeners registered
STDIN-DATA bytes=246 preview="{\"method\":\"initialize\",\"params\":{\"protocolVersion\":\"2025-11-25\",..."
DISPATCH method=initialize id=0
HANDLE-OK id=0 method=initialize, about to writeOutput
STDOUT-WRITE-CALL bytes=153
STDOUT-WRITE-RC flushed=true
HANDLE-WROTE id=0 method=initialize
STDOUT-WRITE-CB ok
STDIN-DATA bytes=55 preview="{\"method\":\"notifications/initialized\",..."
DISPATCH method=notifications/initialized id=null
HANDLE-OK id=null method=notifications/initialized, response=null (notification)
STDIN-DATA bytes=59 preview="{\"method\":\"tools/list\",..."
DISPATCH method=tools/list id=1
HANDLE-OK id=1 method=tools/list, about to writeOutput
STDOUT-WRITE-CALL bytes=21181
STDOUT-WRITE-RC flushed=true
HANDLE-WROTE id=1 method=tools/list
STDOUT-WRITE-CB ok
\`\`\`

Two takeaways:

1. \`require.main\` under Claude Desktop's built-in Node points at \`...\.vite\build\mcp-runtime\nodeHost.js\`, not the bridge. The classic \`if (require.main === module)\` guard evaluates false; without the argv fallback in this PR, \`startBridge()\` never ran and the client sat at 60-second initialize timeout. Posted upstream as [modelcontextprotocol/servers#2729](https://github.com/modelcontextprotocol/servers/issues/2729#issuecomment-4315135297) for visibility.
2. \`STDIN-END\` doesn't actually fire here under Claude Desktop on Windows in this trace, so the previous \`process.exit(0)\` on stdin end wasn't the active bug. Removing it is still correct defensive practice -- Claude's transport may close stdin while waiting for in-flight stdout responses, and exiting would silently drop them.

The protocolVersion negotiation change (commit \`a58983d\`) was a red herring relative to this specific bug -- Claude Desktop accepts \`2025-11-25\` echoed back fine in the trace above. But mirroring the bundled SDK's pinned \`LATEST_PROTOCOL_VERSION='2025-06-18'\` is still spec-defensible and matches what Anthropic's own Filesystem MCP server returns, so I left it in.

Test suite: 264/282 pass, 0 fail, 18 skipped (Thunderbird-running guards). Two test fixes pulled in alongside (test-only changes, no production impact):
- \`test/mcp-bridge.test.cjs:208\` macOS uid filter test now overrides stat.uid for both files (Windows fs.statSync returns uid=0 for everything).
- \`test/tool-access.test.cjs:786\` structural sync tests now scan \`lib/tools-schema.sys.mjs\` and \`lib/constants.sys.mjs\` instead of \`api.js\` after the modular split.

TKasperczyk:
Thanks for the careful module boundaries -- the \`make<Module>({ deps })\` factory pattern is the right shape for breaking up api.js, and a static module-graph validator is the kind of thing you'd only build if you actually planned to live with the result.

Two structural concerns and a few cleanups before this can land:

- **PR #80 just merged.** Your stack is built on top of it, so a rebase should collapse the diff substantially. Worth doing before further review so the actual delta becomes visible.

- **Scope is too broad for a single landing.** The current 12k LoC bundles (a) the module split itself, (b) \`.github/workflows/ci.yml\`, (c) \`eslint.config.mjs\`, (d) Docker integration test harness, (e) \`scripts/check-imports.cjs\` and the tracing infra (\`trace.sys.mjs\`, \`scripts/trace-*\`), (f) JSON pref store. Each of those is a sensible addition individually, but combining them means a regression in any one of them requires reverting the entire refactor.

  What I'd suggest is roughly:
  1. CI workflow + ESLint config (small, lands first, lets all subsequent PRs run CI).
  2. JSON pref store (independent, small).
  3. Tracing infra (independent).
  4. Docker integration tests (independent).
  5. The module split itself (lands last, when the surrounding scaffolding is in place -- and reviewable as a clean refactor).

  That way each piece gets its own focused review and the refactor is reviewable on its own merits.

- **Drop \`.claude/settings.local.json\`.** That's local Claude Code config and shouldn't be in version control. Add it to \`.gitignore\` if useful.

- **Smoke-test phases 3-5 in a running TB.** You flagged in the PR body that those phases aren't yet verified live. There's no CI exercising the extension yet (this PR adds it), so a manual run-through of the compose paths is needed before un-drafting.

Happy to review (1) first if you want to start there -- the CI workflow on its own would already be a win for upstream.

Fuijin:
Thanks for the careful review — the splitting suggestion is sound, and I've been working through it since. Three structural points before I open the first sub-PR, plus a status update on the smoke gap you flagged.

## Counter on "refactor lands last"

The prevailing pattern for refactor-vs-feature merge order is the inverse — refactor lands **first** (or as early as possible after CI scaffolding), features stack on top:

- Kyle Shevlin's [git workflow for refactoring](https://kyleshevlin.com/my-git-workflow-for-refactoring/): *"merge the refactoring branch first… allowing you to easily merge that branch in as well."*
- GitHub's own [stacked-PRs feature](https://github.blog/changelog/) (shipped May 2026) productizes Meta/Google/Uber's decade-old internal pattern with refactor at the stack base + cascading one-click rebase.
- [Graphite's stacking docs](https://graphite.com/guides/advanced-git-branching-strategies) make the same recommendation.

When the refactor lands last, the refactor owner ports every in-flight feature PR onto the new structure manually. That's the anti-pattern.

Concretely for this repo: after rebasing onto your post-#80 main, I had to port 11 commits from #80 / #87 / #90 follow-ups (`fa4bcf8`, `55f0f9c`, `d55527e`, `e7f0855`, `e8dca00`, `cc5e3ca`, `f41a585`, `d25d1d3`, `ef6d683`, `61f3faf`, `5667e08`) from `api.js` to the corresponding `lib/*.sys.mjs` modules. Each was straightforward, but it's roughly an evening of work that wouldn't exist if the refactor had been the base.

Currently open upstream PRs reinforce this: **#89** (bb1, WSL2 unified bridge, +543/-89, rewrites `mcp-bridge.cjs`), **#92** (realitix, attachments inline file paths on bridge, +137), **#93** (realitix, listEvents organizer/attendees/myPartStat), **#94** (realitix, RRULE recurrence), **#95** (realitix, per-occurrence delete/update). If those land before the refactor, that's ~20 more manual ports for me, all touching the same `lib/calendar.sys.mjs` + `lib/compose.sys.mjs` + bridge zones the refactor just modularized.

I'd propose **refactor as #2** (after CI/lint foundation), not #5. Same atomic decomposition, different ordering. Your "focused reviews" framing still applies — the refactor PR just gets reviewed earlier.

## Revised upstream sequence — 6 PRs

After a second pass on what's contributor-toolchain vs my personal dev setup, the upstream sequence is 6 PRs, mapping closely to your original 5:

| # | branch / scope | size | depends on |
|---|---|---|---|
| 1 | `chore/ci-workflow-and-lint` (the OSS-tooling stack — details below) | small-medium (~600 LoC config across 9 files, all config / no runtime) | — |
| 2 | `refactor/api-modular-split` | large but the diff IS the refactor | (1) |
| 3 | `feat/json-pref-store` | small | (1) |
| 4 | `feat/security-permission-engine` (4-layer policy engine + options-page UI) | medium | (3) |
| 5 | `feat/new-mcp-tools` (`inbox_inventory`, `bulk_move_by_query`, `createDrafts` + auto-tool-registry: one file = one tool, zero hand-wiring) | medium | (4) |
| 6 | `chore/static-validators` (`scripts/check-imports.cjs` AST validator + `scripts/madger-chromeutils.cjs` tool-wiring audit — both need `lib/` to exist) | small | (2) |

A few things from my earlier draft are staying on my fork rather than going upstream — they're tied to my local dev / test setup and don't have contributor value without the surrounding scaffolding: structured tracing (`lib/trace.sys.mjs` + `scripts/trace-{write,dump}.cjs`), the greenmail Docker integration tier, and a handful of repo-specific CI config files. I'd rather not ship inert infrastructure into your repo.

I'll start (1) regardless of how the ordering shakes out. It's pure win and gives every subsequent PR a CI baseline to run against. Concretely it brings:

- **ESLint v9 flat config** with per-loader file groups (chrome / WebExt / Node), `no-undef` + `promise/catch-or-return` as the rules that actually catch bugs.
- **commitlint** (`@commitlint/config-conventional`) so PR splits stay grep-able by scope.
- **dependency-cruiser** with `no-circular` + `no-orphans` + `no-extension-from-bridge` boundary rules.
- **knip** for dead-code / unused-deps / unused-exports detection.
- **Dependabot** with two ecosystems (npm + github-actions), weekly grouped with conventional-commit prefixes.
- **release-please** for automated changelog + version bumps reading the commits commitlint enforces. Drops the manual `bump 0.5.0 -> 0.5.1` follow-up.
- **dependency-review-action** blocking PRs that introduce CVE-affected or copyleft deps.
- **`SECURITY.md`** + GitHub private vulnerability reporting policy.
- **Workflow `permissions: contents: read` hardening** on every workflow (least-privilege GITHUB_TOKEN; current default in pre-Feb-2023 repos is read+write everywhere).
- **A `quality.yml` workflow** running ESLint + node:test + commitlint (PR-only) in under 2 minutes — fast feedback on every contributor PR.

Each piece is independently revertable. None changes runtime behavior. All are GitHub-native, free, and benefit every contributor PR — including yours — from the moment they land. Also addresses your `.claude/settings.local.json` note (the `.gitignore` addition + `git rm --cached` is part of (1)).

## On smoke-testing phases 3-5 — status update

You flagged that those weren't verified live. I closed the loop end-to-end against a real headless TB 128.14.0esr (the kind of harness that takes too long to be worth shipping into your CI), and the unit-test count went from 277 to 303 over the same period — most of the new tests are the regression coverage for the security hardening pass mentioned below.

The last hanging assertion was `createDrafts`, and the diagnosis turned out interesting enough to share: TB 128's [JavaScript rewrite of nsMsgSend (Bug 1211292)](https://bugzilla.mozilla.org/show_bug.cgi?id=1211292) changes `_deliverMessage` to early-return after `_mimeDoFcc()` for `SaveAsDraft` mode — `notifyListenerOnStopSending` only fires for `nsMsgDeliverNow`. So `listener.onStopSending` never fires on the SaveAsDraft path on TB 128+; the completion signal is now resolution of the promise that `createAndSendMessage` returns. The pre-refactor code waited on `onStopSending` and timed out at 60s. The fix collapses both completion paths through a single idempotent `settle()` — legacy (C++ nsMsgSend listener) and modern (JS promise) — with timer cancel moved inside `settle` so the race is eliminated rather than guarded. Lives in (2) since it's a one-line completion-path change inside `lib/compose.sys.mjs`.

Worth flagging because this same listener change probably affects other ESR 128 consumers of `nsIMsgSend` for non-`Now` deliver modes (e.g. `nsMsgQueueForLater`, `nsMsgDeliverBackground`, `nsMsgSaveAsTemplate`) that also hit the same early-return.

While the refactor was being made green end-to-end, a security hardening pass closed five real bypass paths in `lib/permissions.sys.mjs` worth calling out because they're each independently exploitable:

1. **`maxBatchSize` bypassed by omitting `limit`.** `bulk_move_by_query` falls back to its tools-schema default (1000) when `limit` is absent — but the policy cap only compared against `typeof args.limit === "number"`, so a missing `limit` skipped the check entirely. Fixed by clamping to `rule.maxBatchSize` when missing or over-cap.
2. **`maxBatchSize` bypassed by stringified JSON arrays.** `deleteMessages` already parses `messageIds` from a JSON string for MCP-client compatibility, but `permissions.sys.mjs` only checked `Array.isArray(args.messageIds)` — so a stringified 10K-element array sailed past the cap. Fixed via an `asArrayOrNull()` normalizer that parses the same way the tool does, applied to all maxBatchSize-protected array args.
3. **Per-folder rules ignored for destination folder.** `bulk_move_by_query` only resolved `args.folderPath` to a folder URI for policy evaluation — a per-folder rule on `imap://server/Archive` did not prevent moving *into* Archive, only operating on it as a source. Fixed by extending `resolveContext` to evaluate both source and destination, denying if either fails.
4. **Unknown `accountId` skipped account-default deny.** `args.accountId: "doesnotexist"` matched no entry in `policy.accounts[...]`, so the per-account default-deny silently didn't apply. Fixed by validating `accountKey` against `MailServices.accounts.accounts` and rejecting unknown keys.
5. **Recursive walks ignored per-folder rules.** `bulk_move_by_query({ folderPath: "INBOX", includeSubfolders: true })` evaluated only the rule for `INBOX` — even if `INBOX/Sensitive` had a deny rule, recursion moved messages from it. Pragmatic fix: when the policy has any per-folder rule for the tool, `includeSubfolders=true` is denied. Full subfolder-rule recursion would be the more permissive alternative if you prefer that direction.

Plus 13 new tests in `test/permissions.test.cjs` covering each path. Worth mentioning here because (4) `feat/security-permission-engine` lands later in the sequence — when it does, these protections are part of the diff.

Other hardening from the same pass: `mcp-bridge.cjs` gained a 16 MiB stdin buffer cap (was unbounded — DoS via `\n`-less input), a 64 MiB HTTP response cap, and in-flight-count observability. `lib/dispatch.sys.mjs` got an own-property guard on the registry handler lookup (`Object.prototype.toString` would otherwise be invocable as a "tool"). `lib/messages.sys.mjs` got HTML5-tolerant script/style end-tag matching (`</script bar>` was previously a sanitizer bypass) and IMAP-safe bulk mutators (DB-only `markRead` doesn't round-trip via STORE). `lib/calendar.sys.mjs` got the same HTML5-tolerant anchor + script sanitizer treatment.

## Concrete next step

I'll open (1) `chore/ci-workflow-and-lint` as its own PR within a day, off your current main. That one's blocked on nothing and lands cleanly regardless of how the rest sequences. Happy to discuss the ordering question on that PR or here — whichever you prefer.

Issue #82, opened 2026-04-26
listEvents omits recurring events — returns only single instances


braklo:
## Summary
`listEvents` returns only single (non-recurring) VEVENTs. Recurring series are not expanded into instances within the requested date range — the recurring meetings are missing from the result entirely, even though Thunderbird itself displays them correctly in the calendar view.

## Reproduction
- Thunderbird-MCP v0.5.0 (commit 16784c8)
- IceWarp CalDAV calendar with multiple weekly/monthly recurring events

```json
{
  "tool": "listEvents",
  "calendarId": "<caldav-calendar-id>",
  "startDate": "2026-04-27T00:00:00.000Z",
  "endDate": "2026-05-03T23:59:59.000Z",
  "maxResults": 500
}
```

## Expected
List includes every event instance (single or recurring) whose start time falls within `[startDate, endDate]` — i.e. all weekly Monday meetings on 2026-04-27, weekly Tuesday meetings on 2026-04-28, etc.

## Actual
Returns only single events (flights, all-day items, one-off meetings). All recurring series are missing from the response. Verified independently via direct CalDAV REPORT (`<C:calendar-query>` with `<C:time-range>`) against the same calendar — the server returns the master VEVENT + RECURRENCE-ID overrides, i.e. the data exists. The MCP appears to filter or simply not expand recurring series before returning.

## Workaround
Bypassing the MCP and querying CalDAV directly via curl REPORT.

## Notes
The TB UI (Calendar tab) renders the recurring instances correctly in the same date range, so the expansion logic is available somewhere in TB. The MCP appears to be reading raw VEVENTs rather than expanding the series.

TKasperczyk:
Resolved by PR #87 (merged in feab2c7) plus a follow-up in ef6d683 that rebalances the Phase 1 / Phase 2 caps and removes the out-of-range master fallback on expansion failure. Please reopen if the IceWarp CalDAV reproduction still shows missing recurring events after upgrading to v0.5.1.

braklo:
[...]

PR #87, opened 2026-05-02
fix(listEvents): expand recurring events via two-phase query

andrei-rizoiu:
## Problem

`listEvents` returns no recurring events for offline-backed calendars
(e.g. OWL/Exchange). The root cause is that `FILTER_OCCURRENCES` does not
work for these calendars: recurring masters are stored with `event_start`
set to their original first-occurrence date, which is typically outside
the queried date range. A date-range query therefore never returns them,
and the manual `getOccurrences` expansion at the call site never runs.

## Fix

Replace the single-phase date-range query with a two-phase approach:

**Phase 1** — `getCalendarItems(calendar, rangeStart, rangeEnd)`:
fetches non-recurring events and modified-occurrence exceptions, whose
`event_start` is inside the range.

**Phase 2** — `getItemsAsArray(FILTER_EVENT, 0, null, null)` (no date
filter): fetches all items including recurring masters, skips
non-recurring ones, then calls `recurrenceInfo.getOccurrences(rangeStart,
rangeEnd, 0)` to expand each master and keep only occurrences that fall
within the queried window.

CalDAV calendars (e.g. Google Calendar) pre-expand recurring occurrences
as individual stored items, so they are captured by Phase 1 without
change.

## Performance

Benchmarked on a local machine against live calendars (5 runs each,
avg reported). Calendar Alpha is an Exchange/OWL-backed calendar
(1 191 cached events, 423 recurring masters). "All calendars" spans
9 calendars totalling ~10 500 cached events.

| Query range | Scope         | Events returned | Latency (avg) |
|-------------|---------------|-----------------|---------------|
| 1 day       | Calendar Alpha | 2              | 39 ms         |
| 1 week      | Calendar Alpha | 38             | 49 ms         |
| 1 month     | Calendar Alpha | 139            | 57 ms         |
| 1 week      | All calendars  | 53             | 125 ms        |
| 1 month     | All calendars  | 183            | 139 ms        |

**Tradeoff:** Phase 2 performs a full-calendar scan on every call —
O(total events in calendar), not O(query range). The range size therefore
has minimal impact on latency (39 ms for 1-day vs 57 ms for 1-month on
the same calendar). The cost is bounded by the local SQLite cache size
(no network round-trip); on the tested calendar the overhead is
acceptable. The Thunderbird calendar API does not expose a
"recurring masters only" filter, so a narrower fetch is not possible
without coupling to internal storage internals.

## Testing

Verified against a live Exchange/OWL-backed calendar:
- A daily recurring event ("Crimson"): 0 occurrences before → 5/5
  returned for a Mon–Fri query.
- A weekly recurring event ("Cobalt"): correctly returned on a
  manually-moved instance date.
- Full week comparison: all events visible in the Thunderbird UI were
  returned by `listEvents` with no missing or extra entries across
  Exchange/OWL, Google Calendar (gdata), and CalDAV calendar types.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -17,6 +17,9 @@ Thumbs.db
 # MCP config (contains local paths)
 .mcp.json
 
+# Local dev notes (agent/developer orientation, not for upstream)
+DEV_NOTES.md
+
 # Node.js
 node_modules/
 package-lock.json
--- a/extension/mcp_server/api.js
+++ b/extension/mcp_server/api.js
@@ -2995,43 +2995,58 @@ var mcpServer = class extends ExtensionCommon.ExtensionAPI {
                 const rangeEnd = cal.dtz.jsDateToDateTime(endJs, cal.dtz.defaultTimezone);
                 const limit = Math.min(Math.max(maxResults || 100, 1), 500);
 
-                // Query with date range and occurrence expansion
+                // Two-phase query to correctly handle recurring events.
+                //
+                // The FILTER_OCCURRENCES flag is intended to make the storage layer
+                // expand recurring masters and return individual occurrences within the
+                // date range. In practice this does not work for offline-backed calendars
+                // (e.g. OWL/Exchange): recurring masters are stored with event_start set
+                // to their original first occurrence date, which is typically outside the
+                // queried range, so a date-range query never returns them and the manual
+                // expansion at the call site never runs.
+                //
+                // Fix — Phase 1: fetch non-recurring events and modified-occurrence
+                // exceptions within the date range (their event_start is inside the
+                // range). Phase 2: fetch ALL recurring masters without a date filter,
+                // then expand each with recurrenceInfo.getOccurrences() and keep only
+                // occurrences that fall within the queried range.
                 const FILTER_EVENT = 1 << 3;
-                const FILTER_OCCURRENCES = 1 << 4;
                 const results = [];
                 for (const calendar of targets) {
-                  let items;
+                  const EVENT_COLLECTION_CAP = limit * 10;
+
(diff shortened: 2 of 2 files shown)


TKasperczyk:
Merged, thanks. The OWL/Exchange root cause (recurring masters stored with `event_start` at the original first occurrence, so date-range queries miss them entirely) is well-diagnosed, and the two-phase approach is the right shape. Benchmarks are reassuring -- 139 ms on 10,500 events is fine for the realistic case.

Followed up in ef6d683 with two small fixes from a code review:

- **Cap rebalanced**: Phase 1 and Phase 2 shared a single `EVENT_COLLECTION_CAP = limit * 10`, so on a busy date range Phase 1 could fill it and silently starve Phase 2 -- the exact bug the PR is meant to fix. Phase 1 is now uncapped (already bounded by the date range), and the cap is renamed `OCCURRENCE_EXPANSION_CAP` with a dedicated counter that only counts emitted occurrences.
- **Phase 2 catch fallback removed**: on `getOccurrences()` failure the catch pushed `formatEvent(item, calendar)` unconditionally, but the master's `event_start` is the original first-occurrence date and is almost certainly outside the queried range. That would surface stale years-old events in every response. Replaced with a `console.warn` and no push.

Minor framing note for future PRs: the "CalDAV calendars pre-expand recurring occurrences" claim is only true for Google -- most CalDAV servers (FastMail, IceWarp, Radicale) send master + RECURRENCE-ID, which Thunderbird's CalDAV provider stores as recurring masters and which Phase 2 then handles. The original reporter (#82) was actually on IceWarp, so the fix lands correctly -- just for a slightly different reason than the description suggests.

Closing #82.

PR #90, opened 2026-05-02
feat(calendar): extend createTask/updateTask with HTML descriptions + add listCategories

andrei-rizoiu:
## Summary

Extends `createTask` with full field support, fixes `updateTask` description
persistence for HTML-bodied tasks, and adds a `listCategories` helper tool.

**Design decision — HTML-first task descriptions:**
Both `createTask` and `updateTask` now write task descriptions via
`descriptionHTML` on the Thunderbird `calITodo`. This is intentional:
OWL maps two distinct Exchange body formats onto the `calITodo` —
HTML bodies → `descriptionHTML`, plain-text bodies → `descriptionText`.
OWL's `UpdateEvent` handler diffs old vs new via `ConvertToOWA()`,
which resolves `Body.Value` as `descriptionHTML || descriptionText`.
Using `setProperty("DESCRIPTION", …)` only updates the plain-text
`descriptionText`, leaving `descriptionHTML` unchanged — OWL sees no
diff and silently skips the Exchange `UpdateItem` call.
HTML is well supported by all clients (Thunderbird, Microsoft To Do,
Outlook), enables clickable links in descriptions, and renders as rich
text in Thunderbird's task editing window.

**New / changed `createTask` parameters:**
- `description` — task body; URLs are auto-linked, and
  `<a href="url">custom text</a>` is also supported
- `priority` — 1=high, 5=normal, 9=low
- `categories` — array of label strings (use `listCategories` first)
- `skipReview` — if `true`, saves via `calendar.addItem()` without dialog

**Fix — `updateTask` description not persisting for HTML-bodied tasks:**
`updateTask` previously used `setProperty("DESCRIPTION", …)` to update
the description. This worked for plain-text tasks but silently failed
for tasks whose body was stored as HTML on Exchange (e.g. any task
created via `createTask`, or edited in a rich-text client). OWL saw no
diff in `descriptionHTML` and skipped the Exchange `UpdateItem` call —
the update appeared to succeed locally but reverted on next sync.
The fix mirrors `createTask`: set `newItem.descriptionHTML` via
`descriptionToHTML()`. After this fix, all MCP-managed task descriptions
are HTML on Exchange regardless of prior state.

**New `listCategories` tool:**
- Returns all category names from `cal.category.fromPrefs()`, sorted
- Use before `createTask`/`updateTask` to get exact case-sensitive names

**Implementation notes:**
- `api.js` runs in `addon_parent` scope and imports `CalTodo` from the
  same `resource:///` ESModule singleton as chrome — passing a
  fully-populated CalTodo (no id) to `createTodoWithDialog` pre-fills
  all fields for user review. The prior `null` argument was incorrect.

## Test plan

- [ ] `listCategories` returns existing category names
- [ ] `createTask` (skipReview: false) opens dialog with all fields pre-filled and URLs clickable
- [ ] `createTask` (skipReview: true) saves directly with description, priority, categories applied
- [ ] `updateTask` description update persists to Exchange for tasks with HTML bodies
- [ ] `updateTask` description update persists to Exchange for tasks with plain-text bodies
- [ ] Descriptions mixing `<a href>` links and bare URLs render as clickable
- [ ] Task descriptions render as rich text in Thunderbird's task editing window


TKasperczyk:
Merged, thanks. The OWL `ConvertToOWA` body-diff analysis is the kind of root-cause reasoning that's hard to come by -- explains the silent-update-failure precisely, and the descriptionHTML fix is the right shape. \`listCategories\` is a nice ergonomic addition too.

Followed up in 61f3faf with a handful of fixes from a code review:

- **\`descriptionToHTML\` safety**: previously threw on \`null\`/\`undefined\` (which \`updateTask(description: null)\` would hit when clearing a description), and preserved user-provided \`<a href>\` tags verbatim. The verbatim path matters because the raw HTML is persisted in iCal \`ALTREP\` and read by external clients (Outlook, mobile, etc.) regardless of TB's render-time sanitization, so \`javascript:\`/\`data:\` hrefs would propagate. Now allows only \`http\`/\`https\`/\`mailto\` schemes, escapes anchor inner text, and strips null bytes from input to prevent sentinel collisions.
- **Priority validation**: schema advertises 1=high, 5=normal, 9=low, but any integer (or non-integer) passed straight through to TB. Both \`createTask\` and \`updateTask\` now validate \`priority\` as an integer in 0-9 and reject otherwise.
- **\`listCategories\` shape**: was returning \`{ categories: [...] }\` while \`listCalendars\` / \`listEvents\` / \`listTasks\` all return bare arrays. Aligned.
- **Test fixture**: \`test/tool-access.test.cjs\` has a hardcoded \`ALL_TOOLS\` list that Structural tests assert against. The PR added \`listCategories\` without updating it -- the suite was failing on the merge commit until I added the entry.

Not addressed: \`dist/thunderbird-mcp.xpi\` was rebundled in the merge against an older state of the codebase. Leaving it in place -- it'll be regenerated when v0.5.1 is cut. Future PRs should ideally exclude \`dist/\` changes and let the maintainer regenerate at release time.

PR #91, opened 2026-05-12
fix(listEvents): correct CLASS_OCCURRENCES filter bit (1<<16, not 1<<4)

realitix:
## Summary

`FILTER_OCCURRENCES` in `listEvents` was wired to `1<<4`, which is actually
`ITEM_FILTER_TYPE_JOURNAL` per `calICalendar.idl`. The correct
`CLASS_OCCURRENCES` bit is `1<<16`. With the wrong bit set:

- Storage/local calendars silently returned master items only (the bogus bit
  was ignored), so recurring events appeared once.
- CalDAV calendars rejected the unexpected filter combination, falling
  through to the manual-expansion catch path — which then called
  `getOccurrences(start, end, 0)` and could return empty for open-ended
  series.

End-user symptom: recurring events (daily stand-up, weekly meetings, etc.)
were missing from `listEvents` results.

## Changes

- `FILTER_OCCURRENCES`: `1 << 4` → `1 << 16` so the provider actually
  expands occurrences.
- The fallback catch logs via `console.warn` instead of swallowing
  silently, so provider-specific issues are diagnosable.
- `getOccurrences(start, end, maxCount)` now receives `EVENT_COLLECTION_CAP`
  as `maxCount` (bounds infinite recurrences like daily stand-up) and falls
  back to the master item if expansion returns empty.

## Test plan

- [x] Storage calendar with a weekly recurring event: `listEvents` returns
      one entry per occurrence in the range.
- [x] CalDAV calendar with a daily open-ended recurrence: `listEvents`
      returns up to `EVENT_COLLECTION_CAP` occurrences and does not hang.
- [x] Non-recurring events still appear once.

### DIFF
--- a/extension/mcp_server/api.js
+++ b/extension/mcp_server/api.js
@@ -2995,9 +2995,11 @@ var mcpServer = class extends ExtensionCommon.ExtensionAPI {
                 const rangeEnd = cal.dtz.jsDateToDateTime(endJs, cal.dtz.defaultTimezone);
                 const limit = Math.min(Math.max(maxResults || 100, 1), 500);
 
-                // Query with date range and occurrence expansion
+                // Query with date range and occurrence expansion.
+                // Filter bits per calICalendar.idl: TYPE_EVENT = 1<<3,
+                // CLASS_OCCURRENCES = 1<<16 (NOT 1<<4 which is TYPE_JOURNAL).
                 const FILTER_EVENT = 1 << 3;
-                const FILTER_OCCURRENCES = 1 << 4;
+                const FILTER_OCCURRENCES = 1 << 16;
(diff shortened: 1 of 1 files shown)


realitix:
Closing: this fix is obsolete after upstream commit feab2c7 (PR #87, "fix(listEvents): expand recurring events via two-phase query"). The new `listEvents` no longer uses `FILTER_OCCURRENCES` at all -- it manually expands all recurring masters in Phase 2 via `recurrenceInfo.getOccurrences(rangeStart, rangeEnd, 0)`, which fixes the same root cause (recurring events missing from results) more thoroughly. The buggy `1 << 4` constant no longer exists in the codebase.

PR #92, opened 2026-05-12
fix(attachments): inline file paths on the bridge side

realitix:
## Summary

Attachments passed to `sendMail` / `replyToMessage` / `forwardMessage` as
**string file paths** (e.g. `"/tmp/foo.pdf"`) were silently dropped when
Thunderbird runs in a sandbox that cannot see the host path. Typical case:
the official Thunderbird snap, which only sees `\$HOME` and removable media —
not `/data/...`, not the host `/tmp`. The extension's
`filePathsToAttachDescs` called `file.exists()`, got `false`, and produced
the warning `"Reply window opened (failed to attach: ...)"` with no file in
the compose window.

Inline base64 objects (`{name, contentType, base64}`) already worked because
the extension writes them to its own private `TmpD` before attaching.

## Fix

The bridge now expands string entries on its side: for every `tools/call`
to `sendMail` / `replyToMessage` / `forwardMessage`, walk the `attachments`
array and replace each string path with a `{name, contentType, base64}`
object read straight off the host filesystem. The extension code path is
unchanged — it always sees inline objects.

- Inline objects supplied by the caller pass through untouched (back-compat
  preserved).
- Reads use `fs.promises` and `Promise.all` so multiple attachments load
  in parallel and don't block the bridge's async message loop.
- Missing path / permission denied / oversize file now surface as a clean
  JSON-RPC `-32602` error instead of a silent warning.
- Size cap is 18 MB raw (~24 MB base64) to stay under the extension's
  existing 25 MB `MAX_BASE64_SIZE` limit.
- No new dependencies — content type guessed from a small built-in
  extension map, falling back to `application/octet-stream`.

## Test plan

- [x] Snap install: `sendMail` with `attachments: ["/tmp/foo.pdf"]` now
      attaches the file instead of silently dropping it.
- [ ] Native install: same call still works (no regression).
- [x] Inline base64 attachments still pass through unchanged.
- [x] Missing path returns a clear JSON-RPC error.
- [x] Multiple attachments load in parallel.

### DIFF
--- a/mcp-bridge.cjs
+++ b/mcp-bridge.cjs
@@ -481,6 +481,124 @@ function discoverConnectionInfo(options = {}) {
   return { data: null, attempts, selectedPath: null };
 }
 
+// Max raw bytes for an attachment read from a path before base64 encoding.
+// Encoded size grows ~33%, so 18 MB raw → ~24 MB base64, staying under the
+// extension's 25 MB MAX_BASE64_SIZE limit.
+const MAX_ATTACHMENT_BYTES = 18 * 1024 * 1024;
+
+// Tools whose `attachments` array may contain string file paths that this
+// bridge resolves on the host filesystem before forwarding. Needed because the
+// Thunderbird snap (and other sandboxed installs) cannot see arbitrary host
+// paths like /data/... or the host's /tmp; passing those paths through to the
+// extension results in silent "failed to attach" warnings since file.exists()
+// returns false inside the sandbox. Reading on the bridge side and shipping
+// inline base64 sidesteps the sandbox entirely.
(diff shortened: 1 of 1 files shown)


TKasperczyk:
Thanks @realitix -- merged in `80f00b3`. String file paths now read on the bridge side and shipped as inline base64, fixing the silent "failed to attach" warnings on snap/flatpak Thunderbird where the extension can't see arbitrary host paths. Clean error mapping (ENOENT/EACCES/oversize) to MCP -32602, and a sensible 18 MB raw cap that stays under the extension's 25 MB base64 limit.

PR #93, opened 2026-05-12
feat(listEvents): expose organizer, attendees, myPartStat

realitix:
## Summary

`formatEvent` in `listEvents` now also exposes the **organizer**, the full
**attendee list**, and the invited identity's **participation status**
(via `cal.itip.getInvitedAttendee`). This complements the VEVENT-level
`STATUS` already exposed upstream (which reflects the organizer's decision)
with the attendee's own PARTSTAT.

## Motivation

Lets MCP clients filter out ghost invites left on CalDAV servers — events
the current user declined (or removed entirely from their own client) but
that the server still serves. Examples:

- Briefing tools that show "what's on my calendar today" can hide events
  the user said no to.
- Calendar sync tools can detect declined recurrences and skip them.

Without `myPartStat`, the client has no way to distinguish "I'm attending"
from "I declined this".

## Shape

Each attendee is reduced to a small dict:

```js
{ id, commonName, participationStatus, role }
```

And every accessor (`item.organizer`, `item.getAttendees()`,
`cal.itip.getInvitedAttendee`) is guarded so calendar providers that
don't expose the helper degrade to empty values rather than failing the
whole `listEvents` call.

## Test plan

- [x] CalDAV calendar with an invitation: `listEvents` returns
      `organizer.id`, an `attendees` array with at least one entry, and
      `myPartStat` set to `ACCEPTED` / `DECLINED` / `TENTATIVE` /
      `NEEDS-ACTION`.
- [x] Event with no organizer/attendees (a personal event): the three
      fields are present with empty/null values.
- [x] Provider that doesn't expose `getAttendees` (older Thunderbird):
      falls back to `attendees: []` without throwing.

### DIFF
--- a/extension/mcp_server/api.js
+++ b/extension/mcp_server/api.js
@@ -4435,6 +4435,25 @@ var mcpServer = class extends ExtensionCommon.ExtensionAPI {
               return VEVENT_STATUS_MAP[String(status).trim().toLowerCase()] || null;
             }
 
+            function formatAttendee(att) {
+              if (!att) return null;
+              return {
+                id: att.id || "",
+                commonName: att.commonName || "",
+                participationStatus: att.participationStatus || "",
+                role: att.role || "",
+              };
+            }
+
+            // The organizer is not an attendee: PARTSTAT and ROLE don't apply.
+            function formatOrganizer(org) {
+              if (!org) return null;
+              return {
+                id: org.id || "",
+                commonName: org.commonName || "",
+              };
+            }
+
             function formatEvent(item, calendar) {
               const allDay = item.startDate ? item.startDate.isDate : false;
               // For all-day events, iCal DTEND is exclusive. Convert to inclusive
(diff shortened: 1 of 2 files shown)


TKasperczyk:
Thanks for this @realitix. A few things before merge:

- `cal.itip.getInvitedAttendee` doesn't look right -- per searchfox, the helper is exposed through calendar scheduling support and called as `getInvitedAttendee(item)`, not under the `cal.itip` namespace. The current code path hits the catch silently, so `myPartStat` always ends up `""` -- which kills the headline feature. Could you cross-check against current comm-central?
- `organizer` is built with the attendee formatter, so it ends up with `participationStatus` and `role` -- those don't apply to organizer (and organizer PARTSTAT can be misleading vs the invited-attendee PARTSTAT). Reducing to `{ id, commonName }` would match the PR description.
- `myPartStat` is camelCase but attendees use `participationStatus`. Maybe rename to `myParticipationStatus` for consistency, or rename attendee field to `partStat` -- either way, one shape.
- A small `formatEvent` test would catch the namespace issue. See `test/empty-trash-compat.test.cjs` for the pattern -- pure JS mirror of the formatter logic, no XPCOM.

Happy to merge once those are sorted.

TKasperczyk:
Re-checked the `cal.itip.getInvitedAttendee` concern against current comm-central and it actually holds up -- scratch that one, my call was wrong. `cal.itip` is defined via `ChromeUtils.defineESModuleGetters(cal, { itip: ".../calItipUtils.sys.mjs" })`, and `getInvitedAttendee(aItem, aCalendar)` is exported two-arg, which is exactly how you call it. The `catch` is graceful degradation, not masking a namespace error -- `myPartStat` should populate fine.

Remaining items before merge, all minor:
- `organizer` is built with `formatAttendee`, so it carries `participationStatus`/`role` that don't apply to an organizer. Reduce it to `{ id, commonName }`.
- `myPartStat` vs the attendees' `participationStatus` -- pick one shape; `myParticipationStatus` would match.
- A small pure-JS `formatEvent` test (see `test/event-online-meeting.test.cjs` for the mirror pattern).

Needs a rebase onto current main too. Once those are in I'll merge.

TKasperczyk:
Gentle nudge @realitix -- still interested in landing this. The three small items from the last review (organizer reduced to `{id, commonName}`, the `myPartStat`/`participationStatus` naming unification, a pure-JS `formatEvent` test) plus a rebase are all that's left. If you're short on time, say the word and I'll pick it up from your branch with credit.

realitix:
Thanks for the review, and sorry for the delay. All three items are in: organizer reduced to `{id, commonName}` (dedicated `formatOrganizer`), `myPartStat` renamed to `myParticipationStatus` to match the attendees' field, and a pure-JS `formatEvent` test (`test/event-attendees.test.cjs`) mirroring `event-online-meeting.test.cjs`. Rebased on current main, CI is green.

Two behaviors you may want to weigh in on: `getInvitedAttendee` returns null on calendars without an iMIP identity, so `myParticipationStatus` degrades to `""` on plain personal calendars (the catch now `console.warn`s on provider throws at least). And attendees are now embedded on every expanded occurrence, which can get heavy for large meetings over long ranges. I kept both as-is, tell me if you want a cap or a different shape.

PR #94, opened 2026-05-12
feat(events): support RRULE recurrence in create/update/list

realitix:
## Summary

Adds RRULE recurrence support to `createEvent`, `updateEvent`, and
`listEvents`.

## Changes

- **`createEvent`** accepts an optional `recurrence` parameter taking an
  iCal RRULE string (e.g. `FREQ=WEEKLY;BYDAY=MO,TU,TH,FR` or
  `RRULE:FREQ=DAILY;COUNT=10`). The `RRULE:` prefix is optional and added
  automatically if missing. Invalid rules are rejected with a clear error.
- **`updateEvent`** accepts the same `recurrence` parameter. Empty string
  or `null` clears the existing recurrence and turns the event into a
  one-shot; a non-empty string replaces the rule.
- **`listEvents`** / `formatEvent` now exposes the RRULE in clear text
  (without the `RRULE:` prefix) under a `recurrence` field on the master
  event, alongside the existing `isRecurring` boolean.

## Implementation notes

- `cal.createRecurrenceInfo()` and `cal.createRecurrenceRule()` are **not**
  exposed on the `cal` namespace in Thunderbird's `calUtils`. The PR uses
  the XPCOM contracts (`@mozilla.org/calendar/recurrence-info;1` and
  `@mozilla.org/calendar/recurrence-rule;1`) like the rest of the codebase
  already does for other `calIxxx` instances.
- Two small helpers were extracted to keep `createEvent` / `updateEvent` /
  `formatEvent` readable:
  - `setRecurrenceOnItem(item, recurrence)`
  - `extractRRuleFromItem(item)`

## Test plan

- [x] `createEvent` with `recurrence: "FREQ=WEEKLY;BYDAY=MO"`: event appears
      every Monday in Thunderbird's calendar UI.
- [x] `createEvent` with `recurrence: "RRULE:FREQ=DAILY;COUNT=5"`: 5
      occurrences only.
- [x] `createEvent` with `recurrence: "GARBAGE"`: returns
      `{ error: "Invalid recurrence rule: ..." }`.
- [x] `updateEvent` with `recurrence: ""`: existing recurring event becomes
      a one-shot.
- [x] `listEvents` on a recurring master returns
      `recurrence: "FREQ=WEEKLY;BYDAY=MO"` (no `RRULE:` prefix) on the
      expanded occurrences.
- [x] Non-recurring events return `recurrence: null`.

### DIFF
--- a/extension/mcp_server/api.js
+++ b/extension/mcp_server/api.js
@@ -1283,7 +1283,7 @@ var mcpServer = class extends ExtensionCommon.ExtensionAPI {
         name: "createEvent",
         group: "calendar", crud: "create",
         title: "Create Event",
-        description: "Create a calendar event through a review dialog. The skipReview safety block is on by default; direct creation is honored only when the user explicitly disables that preference.",
+        description: "Create a calendar event through a review dialog. The skipReview safety block is on by default; direct creation is honored only when the user explicitly disables that preference. Recurring events are supported via the recurrence parameter.",
         inputSchema: {
           type: "object",
           properties: {
@@ -1298,6 +1298,7 @@ var mcpServer = class extends ExtensionCommon.ExtensionAPI {
             showAs: { type: "string", enum: ["busy", "free"], description: "How the event appears in the calendar: 'busy' (solid block, TRANSP:OPAQUE + STATUS:CONFIRMED) or 'free' (hatched, TRANSP:TRANSPARENT + STATUS:TENTATIVE). Defaults to 'busy'. Overridden per-property by explicit status parameter." },
             categories: { type: "array", items: { type: "string" }, description: "Category labels (optional). Category names are case-sensitive; use listCategories to get exact existing names before setting." },
             onlineMeeting: { type: "boolean", description: "If true, generates a Microsoft Teams meeting link via Exchange (OWL/Office 365 accounts only). After creation, OWL embeds the join URL in the event description and exposes it via listEvents (onlineMeetingURL). No-op on non-OWL backends." },
+            recurrence: { type: "string", description: "iCalendar RRULE string for recurring events (e.g. 'FREQ=WEEKLY;BYDAY=MO,TU,TH,FR' or 'RRULE:FREQ=DAILY;COUNT=10'). The 'RRULE:' prefix is optional and added automatically if missing. FREQ=SECONDLY and FREQ=MINUTELY are rejected." },
             skipReview: { type: "boolean", description: "Request direct creation without a review dialog. Honored only when the user explicitly disables the default-on skipReview safety block (default: false)." },
           },
           required: ["title", "startDate"],
@@ -1323,7 +1324,7 @@ var mcpServer = class extends ExtensionCommon.ExtensionAPI {
         name: "updateEvent",
         group: "calendar", crud: "update",
         title: "Update Event",
-        description: "Update an existing calendar event's title, dates, location, or description",
+        description: "Update an existing calendar event's title, dates, location, description, or recurrence rule.",
         inputSchema: {
           type: "object",
           properties: {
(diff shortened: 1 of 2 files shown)


TKasperczyk:
Thanks @realitix. Few items before merge:

- After `feab2c7` `listEvents` expands recurring masters and passes occurrence proxies to `formatEvent`. `extractRRuleFromItem(item)` only reads `item.recurrenceInfo`, but on occurrence proxies the recurrence lives on `item.parentItem.recurrenceInfo` -- expanded occurrences will show `recurrence: null` and `isRecurring: false`. `(item.parentItem || item).recurrenceInfo` fixes it.
- `extractRRuleFromItem` strips the `RRULE:` prefix but doesn't trim the trailing CRLF that `CalRecurrenceRule.icalString` sometimes appends.
- RRULE normalization is thin -- leading/trailing whitespace, lowercase `rrule:` prefix (currently produces invalid `RRULE:rrule:...`), whitespace-only input.
- `SECONDLY` / `MINUTELY` parse via XPCOM but generate no occurrences. Reject upfront with a clear error rather than save a silently-broken event.
- `updateEvent` schema documents clearing on empty string but the implementation/PR body also accepts `null`. Worth aligning.
- `setRecurrenceOnItem` creates a fresh `recurrenceInfo`, discarding existing EXDATEs / modified exceptions. Probably fine semantically (replace rule = fresh series) but worth a note in the tool description.
- A few RRULE tests would help -- `COUNT` vs `UNTIL`, `BYDAY=MO,TU`, `BYMONTHDAY=15`, prefix/no-prefix. See `test/empty-trash-compat.test.cjs` for the pattern.

Holding #95 until this lands since it stacks on top.

TKasperczyk:
Thanks @realitix. Still need the items from the review above plus a rebase onto current main (showAs/categories/onlineMeeting landed since). The one that's a real bug rather than polish:

`listEvents` expands recurring masters and hands occurrence proxies to `formatEvent`, but `extractRRuleFromItem` only reads `item.recurrenceInfo` -- on an occurrence proxy that's null, the rule lives on `item.parentItem.recurrenceInfo`. So `recurrence`/`isRecurring` come back null/false on every expanded occurrence, which is the feature's main path. `(item.parentItem || item).recurrenceInfo` fixes it.

The rest: trim the trailing CRLF off the icalString, tighten RRULE normalization (a lowercase `rrule:` prefix + no trim currently yields `RRULE:rrule:...`), reject `SECONDLY`/`MINUTELY` upfront (they parse but generate no occurrences), align null-vs-empty-string clearing between schema and impl, note that `setRecurrenceOnItem` discards existing EXDATEs/exceptions, and a few RRULE tests. Holding #95 until this lands.

TKasperczyk:
Nudge @realitix -- this one gates #95, so it's the highest-leverage PR of the calendar stack. Outstanding from the review: the `(item.parentItem || item).recurrenceInfo` fix (occurrence proxies from the expansion now on main make the current extraction return null for every occurrence), RRULE normalization + `SECONDLY`/`MINUTELY` rejection, and a rebase -- the branch has drifted to conflicting. Happy to take it over the finish line myself with credit if you'd rather; just say so.

realitix:
Thanks for the offer, but I'd rather take these over the finish line myself. Everything from the review is in: `(item.parentItem || item).recurrenceInfo` for both `recurrence` and `isRecurring`, trailing CRLF trimmed, RRULE normalization handles a lowercase prefix and trims (no more `RRULE:rrule:`), `SECONDLY`/`MINUTELY` rejected upfront, null-vs-empty-string clearing aligned between schema and impl, the EXDATE-discarding behavior of `setRecurrenceOnItem` noted in code and schema, and RRULE tests added (`test/event-rrule.test.cjs`). Rebased on current main (plays nicely with showAs/categories/onlineMeeting), CI green.

Two extra hardenings while I was in there: `normalizeRRule` now rejects rules without a FREQ part, and `FREQ=HOURLY` is rejected on all-day events. Both silently produced never-firing recurrences otherwise.

PR #95, opened 2026-05-12
feat(events): support per-occurrence delete and update via recurrenceId

realitix:
## Summary

Adds **per-occurrence** delete and update to `deleteEvent` / `updateEvent`
via a new `recurrenceId` parameter. Today both tools operate on the whole
series for recurring events — you can't say "cancel only next Monday's
stand-up" or "this one occurrence moves to 2pm".

> ⚠️ **Stacked on #94** (`feat(events): support RRULE recurrence in
> create/update/list`). The diff currently includes that PR's changes; it
> will narrow to just the per-occurrence changes once #94 merges. The
> per-occurrence functionality is independent of #94 at runtime — a
> Thunderbird user can already have RRULE events created elsewhere
> (Thunderbird UI, CalDAV server) — but the helper functions added by #94
> live in the same area of the file, hence the stack.

## Changes

- **`deleteEvent`** accepts an optional `recurrenceId`. When set on a
  recurring master, only that single occurrence is removed via
  `calIRecurrenceInfo.removeOccurrenceAt()` (EXDATE) instead of deleting
  the whole series.
- **`updateEvent`** accepts the same `recurrenceId`. When set, only that
  occurrence is modified via `calIRecurrenceInfo.modifyException()`
  (createException) instead of the master event. Cannot be combined with
  the `recurrence` rule parameter — rules apply to the master only, and
  the call returns a clear error.
- `updateEvent` internals: title/dates/location/description handling is
  extracted into a small `applyEventChanges` helper shared by both the
  master-update and per-occurrence-update paths.
- A `recurrenceIdToCalDateTime(masterItem, recurrenceId)` helper is shared
  by both `deleteEvent` and `updateEvent`.

## Timezone fix

The iCal `RECURRENCE-ID` generated by the rrule engine uses the master's
own timezone, not UTC. Parsing the ISO recurrenceId as UTC caused
`removeOccurrenceAt()` and `getOccurrenceFor()` to silently no-op —
`modifyItem()` succeeded but produced no actual EXDATE. The helper parses
the recurrenceId in `item.startDate.timezone` (falling back to
`defaultTimezone`), which yields the same instant with the correct tz for
the iCal equality check.

## Test plan

- [x] Weekly stand-up: `deleteEvent(eventId, calendarId, recurrenceId)`
      with next Monday's ISO timestamp adds an EXDATE; that one occurrence
      disappears, the rest of the series stays.
- [x] `updateEvent` with `recurrenceId` and a new `startDate`: that one
      occurrence shifts, the master is untouched.
- [x] `updateEvent` with both `recurrence` and `recurrenceId`: rejected
      with a clear error.
- [x] `deleteEvent` with `recurrenceId` on a non-recurring event: error
      `"Event is not recurring; recurrenceId cannot be applied"`.
- [x] `updateEvent` with `recurrenceId` not matching any occurrence: error
      `"No occurrence found at recurrenceId: ..."`.
- [x] CalDAV calendar: changes survive a server round-trip (server-side
      EXDATE / modified instance both appear).

### DIFF
--- a/extension/mcp_server/api.js
+++ b/extension/mcp_server/api.js
@@ -1283,7 +1283,7 @@ var mcpServer = class extends ExtensionCommon.ExtensionAPI {
         name: "createEvent",
         group: "calendar", crud: "create",
         title: "Create Event",
-        description: "Create a calendar event through a review dialog. The skipReview safety block is on by default; direct creation is honored only when the user explicitly disables that preference.",
+        description: "Create a calendar event through a review dialog. The skipReview safety block is on by default; direct creation is honored only when the user explicitly disables that preference. Recurring events are supported via the recurrence parameter.",
         inputSchema: {
           type: "object",
           properties: {
@@ -1298,6 +1298,7 @@ var mcpServer = class extends ExtensionCommon.ExtensionAPI {
             showAs: { type: "string", enum: ["busy", "free"], description: "How the event appears in the calendar: 'busy' (solid block, TRANSP:OPAQUE + STATUS:CONFIRMED) or 'free' (hatched, TRANSP:TRANSPARENT + STATUS:TENTATIVE). Defaults to 'busy'. Overridden per-property by explicit status parameter." },
             categories: { type: "array", items: { type: "string" }, description: "Category labels (optional). Category names are case-sensitive; use listCategories to get exact existing names before setting." },
             onlineMeeting: { type: "boolean", description: "If true, generates a Microsoft Teams meeting link via Exchange (OWL/Office 365 accounts only). After creation, OWL embeds the join URL in the event description and exposes it via listEvents (onlineMeetingURL). No-op on non-OWL backends." },
+            recurrence: { type: "string", description: "iCalendar RRULE string for recurring events (e.g. 'FREQ=WEEKLY;BYDAY=MO,TU,TH,FR' or 'RRULE:FREQ=DAILY;COUNT=10'). The 'RRULE:' prefix is optional and added automatically if missing. FREQ=SECONDLY and FREQ=MINUTELY are rejected." },
             skipReview: { type: "boolean", description: "Request direct creation without a review dialog. Honored only when the user explicitly disables the default-on skipReview safety block (default: false)." },
           },
           required: ["title", "startDate"],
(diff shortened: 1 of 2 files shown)


andrei-rizoiu:
Hi @realitix — thanks for this feature, it works well overall. I tested it against an Exchange calendar via OWL and found a bug in the per-occurrence edit path worth flagging.

**The issue:** `modifyItem(masterClone, oldItem)` is called after `modifyException()`, passing the *master* item. OWL's `modifyItem` detects per-occurrence edits by checking `aNewEvent.recurrenceId` on the item it receives. Since the master has no `recurrenceId`, OWL sends an `ItemId` targeting the whole series rather than an `OccurrenceItemId` — Exchange then overwrites all occurrences on the next sync.

**The fix:** Pass the occurrence directly:

```js
await calendar.modifyItem(modOcc, occurrence);
```

`calICalendar` providers handle occurrence items via `recurrenceId`; this is correct universally. The `modifyException + modifyItem(master)` pattern is a workaround that breaks Exchange occurrence detection.

The fix is in my integration branch if you want to take a look: https://github.com/andrei-rizoiu/thunderbird-mcp/commit/fe5f09c

TKasperczyk:
Holding this until #94 lands -- since the per-occurrence functionality stacks on the RRULE helpers and the diff currently includes #94's changes, easier to review against a clean base. The intrinsic per-occurrence work looks right at a glance, but I want to leave specific feedback against a rebase that drops the #94 overlap. Will revisit as soon as #94 is in.

TKasperczyk:
Holding this until #94 lands -- it stacks on those helpers, easier to review once the base is in. When you rebase to per-occurrence-only: the UPDATE path needs `await calendar.modifyItem(modOcc, occurrence)` rather than passing the master clone, or OWL/Exchange targets the whole series and overwrites every occurrence on the next sync (per @andrei-rizoiu's note above, fork `fe5f09c`). Leave the DELETE path as-is -- `removeOccurrenceAt` + EXDATE is correct there.

realitix:
Rebased on top of #94. The UPDATE path now does `await calendar.modifyItem(modOcc, occurrence)` as requested (thanks @andrei-rizoiu for the fork reference), DELETE unchanged (`removeOccurrenceAt` + EXDATE) with one addition: a `getOccurrenceFor` existence check first, so a recurrenceId that doesn't land on an occurrence errors instead of writing a stray EXDATE and reporting success. The occurrence path also applies status/showAs/categories/onlineMeeting (shared helper with the master path), and all-day recurrenceIds are parsed from the ISO date digits, since `Date` getters shifted the day west of UTC.

PR #100, opened 2026-05-12
Use package version when building manifest

### LINKED ISSUE #99: Build scripts downgrade manifest version from 0.5.1 to 0.5.0 on main
Hi! I noticed a small versioning issue when building from the current main branch.

What happens:

package.json and extension/manifest.json both declare version 0.5.1, but running the build script rewrites extension/manifest.json to 0.5.0.

This appears to happen because the build scripts derive the manifest version from git describe --tags --always, and the latest tag is currently v0.5.0, while main has moved ahead to package version 0.5.1.

Steps to reproduce:

1. Clone the repo:
git clone https://github.com/TKasperczyk/thunderbird-mcp.git

2. Enter the repo:
cd thunderbird-mcp

3. Build the XPI:
node scripts/build-xpi.cjs

4. Check extension/manifest.json.

Actual result:

The manifest version becomes "0.5.0".

Expected result:

The manifest version should remain aligned with package.json, currently "0.5.1".

Impact:

A locally built XPI from main gets packaged with an older extension version than the source package declares. This can make local installs and updates confusing, especially when testing current main.

Possible fix:

Use package.json as the source of truth for the Thunderbird extension manifest version, and keep git describe only for build metadata such as buildinfo.json.

This likely affects both scripts/build-xpi.cjs and scripts/build.sh.

Thanks for the project. I tested this while setting it up with Betterbird, and otherwise it works nicely there too.

emercons:
## Summary

Fixes #99.

This keeps `extension/manifest.json` aligned with `package.json` when building from `main`, while still using `git describe` for build metadata in `buildinfo.json`.

## Changes

- Read the package version from `package.json` in the cross-platform XPI builder.
- Do the same in the shell build script when Node is available.
- Keep the existing git-tag parsing as a fallback if `package.json` cannot be read.

## Verification

- Ran `node scripts/build-xpi.cjs` locally.
- Verified the generated XPI manifest version is `0.5.1` instead of being downgraded to `0.5.0`.

Note: I intentionally did not include the regenerated `dist/thunderbird-mcp.xpi` in this PR to keep the patch focused on the build logic.

### DIFF
--- a/scripts/build-xpi.cjs
+++ b/scripts/build-xpi.cjs
@@ -104,6 +104,16 @@ function addDir(zip, dir, prefix) {
   }
 }
 
+function readPackageVersion() {
+  const packageFile = path.join(PROJECT_DIR, 'package.json');
+  try {
+    const pkg = JSON.parse(fs.readFileSync(packageFile, 'utf8'));
+    return typeof pkg.version === 'string' && pkg.version ? pkg.version : null;
+  } catch {
+    return null;
+  }
+}
+
 // Stamp buildinfo.json with git describe version and timestamp
 const { execSync } = require('child_process');
 const BUILDINFO_FILE = path.join(EXT_DIR, 'buildinfo.json');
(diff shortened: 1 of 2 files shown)


TKasperczyk:
Closing -- `6f967a4` on main already does this (reads `package.json`, writes manifest), with an explicit error on unreadable `package.json` instead of silently falling back to `git describe` (which was the underlying #99 bug). Your writeup on #99 is what got us there, thanks.

PR #105, opened 2026-05-15
feat(calendar): add showAs parameter to createEvent and updateEvent

andrei-rizoiu:
## Summary

- Adds optional `showAs` parameter (`"busy"` | `"free"`) to `createEvent` and `updateEvent`
- `"busy"` (default) — event blocks calendar time and renders as a solid block in Thunderbird
- `"free"` — event does not block calendar time and renders with a distinct (hatched) appearance in Thunderbird
- Explicit `status` parameter is independent and always takes precedence

## Motivation

Before this change, events created via the MCP had no `TRANSP` property set. Without it, some calendar backends (notably Exchange/OWL) default the event to `LegacyFreeBusyStatus: Free`, which after sync writes `TRANSP:TRANSPARENT` back to the local store. The result: all MCP-created events appeared as "Free" and did not block calendar time.

The `showAs` parameter gives callers explicit control, with a safe default of `"busy"` so new events block time as expected.

## Behaviour notes

- Thunderbird's calendar rendering uses two orthogonal signals: `TRANSP` controls free/busy status for scheduling (visible in OWA/CalDAV), while a separate internal signal controls the visual appearance in the Thunderbird UI. `showAs` sets both consistently so the event looks and behaves correctly on all platforms.
- Tested on Exchange (OWL) and Google CalDAV (DIN) calendars.
- `updateEvent` with `showAs: null` clears the scheduling signal only.

## Test plan

- [ ] `createEvent` without `showAs` → event blocks time (Busy) by default
- [ ] `createEvent` with `showAs: "busy"` → solid block, blocks calendar time
- [ ] `createEvent` with `showAs: "free"` → distinct appearance, does not block calendar time
- [ ] `updateEvent` toggling `showAs` → updates correctly in both Thunderbird and OWA/CalDAV
- [ ] Verify on both CalDAV and Exchange-backed calendars

### DIFF
--- a/extension/mcp_server/api.js
+++ b/extension/mcp_server/api.js
@@ -272,6 +272,7 @@ var mcpServer = class extends ExtensionCommon.ExtensionAPI {
             calendarId: { type: "string", description: "Target calendar ID (from listCalendars, defaults to first writable calendar)" },
             allDay: { type: "boolean", description: "Create an all-day event (default: false)" },
             status: { type: "string", description: "VEVENT STATUS: 'tentative', 'confirmed', or 'cancelled'. Defaults to confirmed if omitted." },
+            showAs: { type: "string", enum: ["busy", "free"], description: "How the event appears in the calendar: 'busy' (blocks time, TRANSP:OPAQUE) or 'free' (TRANSP:TRANSPARENT). Defaults to 'busy'." },
             categories: { type: "array", items: { type: "string" }, description: "Category labels (optional). Category names are case-sensitive; use listCategories to get exact existing names before setting." },
             skipReview: { type: "boolean", description: "If true, add the event directly without opening a review dialog (default: false)" },
           },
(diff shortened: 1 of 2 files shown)


TKasperczyk:
Thanks @andrei-rizoiu -- the OWL/Exchange free/busy default is a real footgun. Few items:

- `createEvent` doesn't validate `showAs` (any string passes; only `"free"` is special-cased, everything else falls through to busy). `updateEvent` rejects invalid values, so the two are inconsistent. Mirror the `updateEvent` enum check in `createEvent`.
- The "separate internal signal" you mention is actually `STATUS` (TENTATIVE for free, CONFIRMED for busy) -- standard iCal, not internal. Worth being explicit in the PR/docs because STATUS syncs to CalDAV/Google/Exchange as real event status, not just visual rendering. Callers passing `showAs: "free"` get tentative status as a side effect unless they pass explicit `status`.
- "Explicit `status` always takes precedence" -- as implemented `status` only overrides the derived STATUS, not TRANSP. So `{ status: "confirmed", showAs: "free" }` produces `STATUS:CONFIRMED` + `TRANSP:TRANSPARENT`. That's actually the right design (orthogonal axes in iCal), but the tool description should say so explicitly.
- `updateEvent` with `showAs: null` clears TRANSP but leaves STATUS at whatever a previous update set. Documenting helps.
- Defaulting new events to busy is a semantic change from current behavior (no TRANSP -> backends default to free). Right default, worth a changelog mention.
- A test for the create-side enum validation + the TRANSP clear path on update would round it out (see `test/empty-trash-compat.test.cjs` for the pattern).

#108 (categories) and #110 (attendees) both touch the same `createEvent`/`updateEvent` signature -- whichever lands second will need a small rebase. Worth coordinating an options-object refactor once these three land, so we stop appending positional args.

TKasperczyk:
Merged in `e8cfede`, with a small fixup folded into the squash and a rebase onto current main (which now includes #108):

- Added the upfront `showAs` enum check to `createEvent` to match `updateEvent`.
- Added `test/event-show-as.test.cjs` -- 12 cases covering create-side validation + STATUS derivation and the update-side TRANSP action tree.
- Briefly tried tightening the updateEvent schema to `type: ["string", "null"]` for null-clearing, but the project's validator only handles scalar `type` -- it would have blocked all string values. Reverted; null still passes through because the validator skips null on optionals, same as your original PR.

Thanks.

PR #108, opened 2026-05-15
feat(calendar): add categories to createEvent, updateEvent, and listEvents

andrei-rizoiu:
## Summary

- `createEvent` and `updateEvent` now accept an optional `categories` array
- `listEvents` now returns a `categories` field on every event
- Pass an empty array to `updateEvent` to clear all categories
- Use `listCategories` to get exact category names before setting (names are case-sensitive)

## Merge order note

This PR touches the same lines as PR #105 (`showAs` parameter). If #105 merges first, this branch will need a rebase before it can merge cleanly. The conflicts are trivial — four lines where both PRs extend the same function signatures and dispatch calls:

```js
// After rebasing on the merged main, these become:
async function createEvent(..., status, showAs, categories)
async function updateEvent(..., status, showAs, categories)
return await createEvent(..., args.status, args.showAs, args.categories)
return await updateEvent(..., args.status, args.showAs, args.categories)
```

Happy to rebase once #105 is merged if that helps.

## Test plan

- [ ] `createEvent` with `categories: ["Personal"]` → category visible in Thunderbird
- [ ] `createEvent` without `categories` → `categories: []` in `listEvents` response
- [ ] `updateEvent` with new category → reflected in `listEvents`
- [ ] `updateEvent` with `categories: []` → clears all categories
- [ ] `listEvents` returns `categories` array for all events

### DIFF
--- a/extension/mcp_server/api.js
+++ b/extension/mcp_server/api.js
@@ -272,7 +272,7 @@ var mcpServer = class extends ExtensionCommon.ExtensionAPI {
             calendarId: { type: "string", description: "Target calendar ID (from listCalendars, defaults to first writable calendar)" },
             allDay: { type: "boolean", description: "Create an all-day event (default: false)" },
             status: { type: "string", description: "VEVENT STATUS: 'tentative', 'confirmed', or 'cancelled'. Defaults to confirmed if omitted." },
-            categories: { type: "array", items: { type: "string" }, description: "Category labels (optional). Use listCategories to get exact existing names before setting." },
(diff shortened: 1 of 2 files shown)


TKasperczyk:
Merged in `bd9914e`, with a small fixup folded into the squash and a rebase onto current main:

- `updateEvent` with `categories: null` was silently clearing all categories because the generic validator allows null on optional fields. Restricted clearing to an explicit `[]` via `Array.isArray(categories)`. `null` / `undefined` now preserve existing categories.
- Added "category names are case-sensitive" to the `createEvent` / `updateEvent` schema descriptions for discoverability (matches what `listCategories` already says).
- Added `test/event-categories.test.cjs` for the null/undefined/empty/non-empty decision tree.

Thanks.

PR #109, opened 2026-05-15
feat(calendar): add onlineMeeting param to createEvent and updateEvent

andrei-rizoiu:
## Summary

- Adds an optional `onlineMeeting` boolean parameter to `createEvent` and `updateEvent`
- When `true`, sets `X-ONLINE-MEETING-PROVIDER: TeamsForBusiness` on the iCal event
- OWL (Owl for Exchange) reads this property via `Event2JSON` and routes to `CreateTeamsMeeting()` in its OWA backend, which asks Exchange/Office 365 to generate a Teams join URL server-side
- The join URL is written back immediately to `X-MICROSOFT-SKYPETEAMSMEETINGURL` and surfaced as `onlineMeetingURL` in `listEvents` responses
- `updateEvent` with `onlineMeeting: false` removes an existing Teams link
- Defaults to `false` — no Teams link is created unless explicitly requested

## Notes

This feature currently works only with OWL/Office 365 Exchange accounts. The `X-ONLINE-MEETING-PROVIDER` property is OWL-specific; other calendar backends (CalDAV, Google) will silently ignore it. In principle the approach could be extended to other providers — for example, Google Meet links could be requested via the Google Calendar API using `conferenceData.createRequest` — but that would require provider-specific handling beyond the scope of this PR.

I considered not submitting this upstream for that reason, but thought it worth leaving to the maintainer's discretion — it adds no overhead for non-OWL users and may be useful for others on Exchange.

## Test plan

- [ ] `createEvent` with `onlineMeeting: true` → event appears in Thunderbird with Teams join link in description and location
- [ ] `listEvents` response includes `onlineMeetingURL` field populated with the join URL
- [ ] `createEvent` without `onlineMeeting` → no Teams link added (default behaviour unchanged)
- [ ] `updateEvent` with `onlineMeeting: true` on existing event → Teams link added
- [ ] `updateEvent` with `onlineMeeting: false` on event with Teams link → link removed

### DIFF
--- a/extension/mcp_server/api.js
+++ b/extension/mcp_server/api.js
@@ -328,6 +328,7 @@ var mcpServer = class extends ExtensionCommon.ExtensionAPI {
             status: { type: "string", description: "VEVENT STATUS: 'tentative', 'confirmed', or 'cancelled'. Defaults to confirmed if omitted." },
             showAs: { type: "string", enum: ["busy", "free"], description: "How the event appears in the calendar: 'busy' (solid block, TRANSP:OPAQUE + STATUS:CONFIRMED) or 'free' (hatched, TRANSP:TRANSPARENT + STATUS:TENTATIVE). Defaults to 'busy'. Overridden per-property by explicit status parameter." },
             categories: { type: "array", items: { type: "string" }, description: "Category labels (optional). Category names are case-sensitive; use listCategories to get exact existing names before setting." },
+            onlineMeeting: { type: "boolean", description: "If true, generates a Microsoft Teams meeting link via Exchange (OWL/Office 365 accounts only). The join URL is returned in the response and embedded in the event description." },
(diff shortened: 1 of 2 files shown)


TKasperczyk:
Merged in `977d15f`, with a small fixup folded into the squash and a rebase onto current main (which now includes #105, #108, #118):

- Conflict resolution: integrated `onlineMeeting` alongside the recently-merged showAs/categories params. Final positional order on both `createEvent` and `updateEvent` is `(..., status, showAs, categories, onlineMeeting)`.
- Schema description tweak: the `createEvent` description said "the join URL is returned in the response", but `createEvent` only returns `{success, message}`; the URL surfaces via `listEvents.onlineMeetingURL` after creation. Updated to reflect that and to call out the OWL/Office-365-only constraint up front.
- Added `test/event-online-meeting.test.cjs` covering the create/update decision tree and the `formatEvent` mapping.

Thanks.

PR #110, opened 2026-05-15
feat(calendar): add attendees param to createEvent and updateEvent

andrei-rizoiu:
## Summary

Adds an optional `attendees` array parameter to `createEvent` and `updateEvent`.

Each attendee entry: `{ email (required), name (optional), role: "required" | "optional" }`.

**createEvent** — attendees are added via `event.addAttendee()`. The event organizer is also set automatically from the calendar's iMIP identity, which is required on Exchange/OWL calendars: OWL's `addItem` checks that `event.organizer.id` matches the calendar's `organizerId` property to distinguish a new meeting (organizer = us) from a Lightning invitation being moved between calendars (unsupported by Exchange). Without the organizer being set, OWL silently strips all attendees before sending to Exchange.

**updateEvent** — replaces the full attendee list. Removes all existing attendees then adds the new list. Pass an empty array to remove all attendees.

On Exchange/OWL calendars, Exchange automatically sends meeting invitation emails to all attendees when the event is created or updated (controlled by `SendMeetingInvitations`/`SendMeetingInvitationsOrCancellations` in EWS, driven by OWL's existing `getNotificationStatus()` mechanism).

## Test plan

- [x] Create event with `attendees: [{email, name, role}]` on an Exchange/OWL calendar → attendee appears in Thunderbird, OWA shows attendee, invite email received
- [x] Create event with `role: "optional"` → attendee has `OPT-PARTICIPANT` role
- [x] `updateEvent` with new attendees list → replaces existing attendees, Exchange sends update notification
- [x] `updateEvent` with `attendees: []` → removes all attendees
- [x] Create event without attendees → behaviour unchanged
- [x] `updateEvent` without attendees param → behaviour unchanged

🤖 Generated with [Claude Code](https://claude.ai/claude-code)

### DIFF
--- a/extension/mcp_server/api.js
+++ b/extension/mcp_server/api.js
@@ -273,6 +273,7 @@ var mcpServer = class extends ExtensionCommon.ExtensionAPI {
             allDay: { type: "boolean", description: "Create an all-day event (default: false)" },
             status: { type: "string", description: "VEVENT STATUS: 'tentative', 'confirmed', or 'cancelled'. Defaults to confirmed if omitted." },
             skipReview: { type: "boolean", description: "If true, add the event directly without opening a review dialog (default: false)" },
+            attendees: { type: "array", items: { type: "object", properties: { email: { type: "string", description: "Attendee email address" }, name: { type: "string", description: "Display name (optional)" }, role: { type: "string", enum: ["required", "optional"], description: "Defaults to required" } }, required: ["email"] }, description: "List of attendees to invite (optional). Supported on Exchange/OWL calendars — Exchange sends meeting invitation emails automatically." },
           },
           required: ["title", "startDate"],
         },
@@ -309,6 +310,7 @@ var mcpServer = class extends ExtensionCommon.ExtensionAPI {
             location: { type: "string", description: "New event location (optional)" },
             description: { type: "string", description: "New event description (optional)" },
             status: { type: "string", description: "New VEVENT STATUS: 'tentative', 'confirmed', or 'cancelled' (optional)" },
+            attendees: { type: "array", items: { type: "object", properties: { email: { type: "string", description: "Attendee email address" }, name: { type: "string", description: "Display name (optional)" }, role: { type: "string", enum: ["required", "optional"], description: "Defaults to required" } }, required: ["email"] }, description: "Replaces the full attendee list. Pass an empty array to remove all attendees. Supported on Exchange/OWL calendars." },
           },
           required: ["eventId", "calendarId"],
         },
@@ -934,6 +936,7 @@ var mcpServer = class extends ExtensionCommon.ExtensionAPI {
             let cal = null;
             let CalEvent = null;
             let CalTodo = null;
+            let CalAttendee = null;
             try {
               const calModule = ChromeUtils.importESModule(
                 "resource:///modules/calendar/calUtils.sys.mjs"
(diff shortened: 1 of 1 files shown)


TKasperczyk:
Thanks @andrei-rizoiu. A few items before merge:

- `updateEvent` with `attendees: null` silently removes everyone -- the validator skips null on optionals and `(attendees || [])` treats null like an empty array. `undefined` means "no change", `[]` means "remove all", `null` should probably no-op. On Exchange/OWL this triggers update/cancellation invitations, which is a destructive surprise.
- `updateEvent` doesn't apply the organizer-from-calendar-identity logic when adding attendees. `createEvent` does. If an existing non-meeting event gets attendees added without an organizer present, OWL silently strips them. The two paths should share the same organizer helper.
- The nested attendee schema is advertised but the generic validator doesn't enforce `items.required.email` or `role enum`, so invalid `email` throws and invalid `role` silently becomes `REQ-PARTICIPANT`. Either accept that the schema is descriptive only, or extend the validator.
- Send-invitations control: on OWL/Exchange this creates a new silent-email-dispatch path with no caller-side control. `blockSkipReview` only covers mail compose. Not blocking this PR but worth tracking -- a follow-up with a `sendInvitations` param or extended `blockSkipReview` coverage would close the gap.
- Organizer setup is reasonable for scheduling events but it's not OWL-only -- any calendar with `imip.identity.key` and attendees gets an `ORGANIZER`, and if `organizerId` is missing the PR sets it on the calendar. Helpful for broken CalDAV configs but it's a silent property mutation, worth flagging in the docs.
- A small decision-tree test for the null/empty/non-empty paths (and the role enum mapping) would help. See `test/empty-trash-compat.test.cjs` for the pattern.

Happy to revisit once those are sorted.

TKasperczyk:
Thanks @andrei-rizoiu. Still the items from the review above -- two are the blockers, plus a rebase onto current main:

- `updateEvent` with `attendees: null` wipes everyone: `(null || [])` collapses to `[]` after the unconditional remove loop. `null` should no-op, `undefined` = no change, `[]` = remove all. Destructive on Exchange/OWL since it fires cancellations.
- `updateEvent` doesn't set the organizer-from-calendar-identity that `createEvent` does, so OWL silently strips the added attendees. Pull that into a shared helper both paths call.

Enum enforcement on the attendee schema and the docs note on the organizerId mutation can follow. Coordinate with #93 on the read side -- same attendee/organizer shaping.

TKasperczyk:
Nudge @andrei-rizoiu -- still want this. The two blockers from the review stand: `attendees: null` currently wipes the whole list via `(null || [])` (destructive on Exchange -- it fires cancellations; null should no-op), and `updateEvent` needs the same organizer-from-identity treatment as create so OWL doesn't strip the added attendees. Plus a rebase; the branch went conflicting under main's calendar changes. As with #93/#94: if you're out of bandwidth, I can finish it from your branch with credit.

andrei-rizoiu:
Thanks for the offer, and sorry for the drag on this one — I'm out of bandwidth to get the null-handling and updateEvent organizer fixes done properly, so please do take it over from `feat/events-attendees` with credit.

andrei-rizoiu:
Quick update so you don't duplicate work: while rebuilding my local deployment on v0.7.4 today (TB 153 forced the refresh), I ended up reapplying this feature with both review blockers fixed — `attendees: null` is now a no-op on updateEvent (only `[]` clears), and a shared `ensureOrganizer()` helper runs on both create and update paths. It's on my fork as [`feat/integration-v074` @ `4862964`](https://github.com/andrei-rizoiu/thunderbird-mcp/commit/4862964). Caveat: it's stacked on #93 + #95 (it reuses `applyEventMetaChanges` from #95's refactor), so it won't apply to plain main until those land.

That said, this was a quick reapply to unblock my own setup — I don't have the bandwidth for comprehensive testing (decision-tree tests, the role enum mapping, Exchange invite/cancellation behaviour), so my hand-over offer stands: I'd still prefer you properly finish it with credit, using my branch as a basis or just as a reference for the conflict resolution — whichever is easier.

PR #113, opened 2026-05-20
In listEvents, return event attendees

frmichel:
Hi @TKasperczyk, in the listEvent serice, I've made a minor change to return the attendees -- if any -- as part of an event. Handy for my use case. In case you'd like to include this into the next version.


frmichel:
Jus saw that there already exists PR #93 about the same feature. I should have checked before, sorry for the duplicate.

TKasperczyk:
Closing in favor of #93, which covers the same area more comprehensively (organizer + attendees + myPartStat, plus provider guards around organizer/attendee extraction and invited-attendee lookup). Thanks @frmichel.

PR #125, opened 2026-05-27
fix(security): tighten default policies and input validation on the outbound-tool surface

JordanRO2:
The originally-scoped six hardenings + their 17 unit tests, rebased onto current `main` (v0.6.0, `1db179f`). The three extra hardenings and the `test/poc/` harness moved to #127 per review.

## The six fixes
- `blockSkipReview` now defaults to `true`; `createEvent` and `createTask` honor the same review gate the mail tools already do.
- File-path attachments run through a cross-platform deny-list (credential stores, system dirs, browser profile stores, Thunderbird's own profile) with a 50 MB cap.
- Filter `buildTerms` / `buildActions` use a strict `hasOwnProperty` allowlist instead of the `parseInt` fallback.
- `validateToolArgs` enforces `enum`, `oneOf`, and nested `additionalProperties` via a small recursive walker (`validateAgainstSchema`).
- `writeConnectionInfo` tightens the tmp-dir mode to `0700` on POSIX.
- 17 new unit tests.

## Rebase note (v0.6.0)
Rebased cleanly onto v0.6.0; the one merge point was `validateToolArgs`, where v0.6.0 added `minItems`/`maxItems` for the new `getMessages` batch tool. Resolved by delegating to the recursive `validateAgainstSchema` (this PR) **and** keeping the array-length bounds (v0.6.0), so getMessages-batch caps stay enforced — covered by both the new validator tests and the existing getMessages-batch tests.

## Tests
`test/validation.test.cjs` **64/64**; full suite green except the one macOS-only `mcp-bridge` test that already fails on non-macOS runners. Source-only (no `dist/` xpi), matching your release-time rebuild cadence.

### DIFF
--- a/extension/mcp_server/api.js
+++ b/extension/mcp_server/api.js
@@ -328,8 +328,9 @@ var mcpServer = class extends ExtensionCommon.ExtensionAPI {
                       name: { type: "string", description: "Attachment filename" },
                       contentType: { type: "string", description: "MIME type, e.g. application/pdf" },
                       base64: { type: "string", description: "Base64-encoded file content" },
+                      content: { type: "string", description: "Alias for base64 (accepted for backwards compatibility); base64 takes precedence when both are set" },
                     },
-                    required: ["name", "base64"],
+                    required: ["name"],
                     additionalProperties: false,
                   },
                 ],
@@ -366,8 +367,9 @@ var mcpServer = class extends ExtensionCommon.ExtensionAPI {
                       name: { type: "string", description: "Attachment filename" },
                       contentType: { type: "string", description: "MIME type, e.g. application/pdf" },
                       base64: { type: "string", description: "Base64-encoded file content" },
+                      content: { type: "string", description: "Alias for base64 (accepted for backwards compatibility); base64 takes precedence when both are set" },
                     },
-                    required: ["name", "base64"],
+                    required: ["name"],
                     additionalProperties: false,
                   },
                 ],
@@ -614,8 +616,9 @@ var mcpServer = class extends ExtensionCommon.ExtensionAPI {
                       name: { type: "string", description: "Attachment filename" },
                       contentType: { type: "string", description: "MIME type, e.g. application/pdf" },
                       base64: { type: "string", description: "Base64-encoded file content" },
+                      content: { type: "string", description: "Alias for base64 (accepted for backwards compatibility); base64 takes precedence when both are set" },
                     },
-                    required: ["name", "base64"],
+                    required: ["name"],
                     additionalProperties: false,
                   },
                 ],
@@ -654,8 +657,9 @@ var mcpServer = class extends ExtensionCommon.ExtensionAPI {
                       name: { type: "string", description: "Attachment filename" },
                       contentType: { type: "string", description: "MIME type, e.g. application/pdf" },
                       base64: { type: "string", description: "Base64-encoded file content" },
+                      content: { type: "string", description: "Alias for base64 (accepted for backwards compatibility); base64 takes precedence when both are set" },
                     },
-                    required: ["name", "base64"],
+                    required: ["name"],
                     additionalProperties: false,
                   },
                 ],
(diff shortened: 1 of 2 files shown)


TKasperczyk:
Thanks for the clean split @JordanRO2. Please move the three extra hardenings (filter forward/reply blocking + injected-header sanitization + untrusted-body wrapping + compose audit-log, contact-write review gate + system-principal scheme allowlist, htmlToMarkdown URL/text sanitization) and the `test/poc/` harness to a follow-up PR.

That brings #125 to the originally-scoped five hardenings + 17 tests, reviewable in one sitting. The extra three are all reasonable but each warrants focused review on its own, especially the htmlToMarkdown URL-scheme work (potential rendering side effects worth examining carefully).

Once #125 is the trimmed version, I'll run a codex review pass and merge.

TKasperczyk:
Thanks @JordanRO2. Ran the codex review pass on the trimmed version -- one blocker before merge.

The attachment validation now rejects a shape the runtime still accepts. `validateToolArgs` runs before dispatch, and the attachment schema on `sendMail` / `saveDraft` / `replyToMessage` / `forwardMessage` is `required: ["name", "base64"]` with `additionalProperties: false`. But the runtime still reads `entry.base64 || entry.content` when building the attachment, so `{ name, content }` -- a previously-valid inline attachment -- now fails validation before it reaches the handler. Callers passing `content` would start breaking.

Fix: accept `content` as a `base64` alias in the attachment schema (keep `contentType` optional, it already is at runtime), or drop the alias deliberately with a note. Worth a regression test for `{name, content}` and for `{name, base64}` without `contentType`.

Two more to fold in while you're here, neither blocking:
- `isSensitiveFilePath` checks the raw path string only -- a benign `/tmp/report.pdf` symlinked to `~/.ssh/id_rsa` slips through. Resolve the `nsIFile` (or reject symlinks) and re-check before attaching.
- `validateAgainstSchema` returns early on `value === null`, so `attachments: [null]` skips the item schema. Minor, but nullability should be explicit per field.

Also needs a rebase onto current main -- #129/#128/#132 just landed. Once the alias is back I'll merge.

JordanRO2:
Thanks for the detailed review. The blocker is fixed and the two follow-ups are folded in:

- **`content` alias (blocker):** the attachment schema on `sendMail` / `saveDraft` / `replyToMessage` / `forwardMessage` now accepts `content` as an alias for `base64` (required relaxed to `["name"]`), so the previously-valid `{name, content}` inline attachment passes validation again. `base64` still takes precedence at runtime when both are set, matching the existing `entry.base64 || entry.content`.
- **Symlink resolution:** `isSensitiveFilePath` now re-checks the canonical path — the `nsIFile` is `normalize()`d (resolving symlinks + `.`/`..`) and the deny-list re-run before the file is read, so a benign path symlinked to a credential store is blocked.
- **Explicit nullability:** `validateAgainstSchema` now rejects `null`/`undefined` array items instead of returning early, so `attachments: [null]` no longer skips the item schema.

Added 6 regression tests (content alias, `base64` without `contentType`, both set, missing name, unknown property, null item) and rebased onto current `main`. Full suite green; `version-sync`, `lint`, and tests on Node 18/22/24 all pass.

TKasperczyk:
Merged in ae9930e -- thanks @JordanRO2 for closing every loop from the last round: the `content` alias fix, the `required` relaxation, the symlink `normalize()` re-check, the explicit null rejection, and the six regression tests all verified good, and the branch merged clean against a main that had moved 15 commits under it.

A post-merge review pass found gaps in the new hardening layer (none of them regressions -- main had no deny-list at all before this), so I followed up on main with a series of fixes:

1. **8c1cace -- inline attachment payload enforcement**: `required: ["name"]` also accepted `{name}` / empty payloads that the runtime silently dropped, and malformed base64 was stripped-to-corrupt instead of rejected. Validation now requires a non-empty payload and strict base64.
2. **9bd1a00 -- Linux profile deny-list layout**: the pattern expected `.thunderbird/profile(s)/`, but real profiles live directly under `~/.thunderbird/<profile>` with `profiles.ini` beside them -- so the deny-list missed the actual Linux layout.
3. **95b8642 -- fail-closed filesystem checks**: normalization/stat errors used to fall through open (size errors coerced to 0 and passed the cap); now they reject. Added a regular-file requirement (no FIFOs/devices) plus aggregate 50MB and 20-attachment per-message caps -- the per-file cap alone allowed multi-GB messages.
4. **e139eef -- the big one, bridge-side enforcement**: the stdio bridge inlines `filePathsToAttach` to base64 *before* forwarding, so the extension-side deny-list never saw path strings on the primary transport at all. The full policy (deny-list, size/count caps, regular-file, symlink rejection, sequential bounded reads) now runs in the bridge too, kept in sync with the extension's copy.
5. **17ebc8b -- symlinked attachment paths are now denied outright**: Windows `nsIFile.normalize()` only normalizes lexically -- it does not resolve junctions -- so the "resolve then re-check" strategy didn't hold there. Links are rejected fail-closed on both sides instead.
6. **d673e0a -- validation now honors the documented base64-over-content precedence** and requires non-empty attachment names.

Plus housekeeping the review surfaced: e4ac2a6 (a pre-existing `removeConnectionInfo` scoping bug your new rejection path could reach), 6e51ccd (MCP notifications now return 202 per spec), and 4157e06/907c012 (README, options UI, and the MCP-visible tool descriptions now all state the default-on review block, including the `createEvent`/`createTask` gating this PR added).

The `blockSkipReview` default flip is the most user-visible change in the batch and it's the right default -- thanks for pushing the whole surface forward. Not in any release yet; one is coming shortly with this plus today's other fixes.

PR #127, opened 2026-05-27
fix(security): three extra hardenings (filter forward/reply, contact-write gate, htmlToMarkdown) + PoC harness

JordanRO2:
Split out of #125 per review: the three extra hardenings (and the `test/poc/` harness) that warrant their own focused review, rebased onto current `main` (v0.6.0, `1db179f`).

**Stacked on #125.** Until #125 merges, this PR's diff includes #125's commit; once it lands, GitHub narrows this to the three commits below.

## The three hardenings (each its own commit)
1. **`fix(security): block filter forward/reply, sanitize headers, wrap untrusted bodies, audit-log compose`** — filters run on every incoming message with no review UI, so a single `createFilter` with a forward/reply action installs a permanent silent-exfiltration rule. New `blockFilterForwardReply` pref (default on) refuses those action types via the MCP API; header values are stripped of CR/LF/NUL; untrusted bodies are wrapped; compose calls are audit-logged.
2. **`fix(security): gate contact writes; allow-list system-principal fetch schemes`** — contact create/update/delete go behind the review pref; system-principal fetches are restricted to an allow-list of mail-store schemes (`mailbox`/`imap`/`news`…), refusing `file://`/`chrome://` etc.
3. **`fix(security): sanitize htmlToMarkdown URL schemes and link text`** — allow-list of safe href schemes, `data:` limited to `data:image/*`, and link-text escaping that defeats the `click](javascript:bad)` markdown-injection. **This is the one worth examining closely for rendering side effects** — it changes how links/images render in the markdown conversion.

Plus `test/poc/` — a small F1 (filter-forward) / F3 (contact-spoof) PoC harness demonstrating the first two.

## Tests
Full suite **429 pass / 17 skipped / 1 fail**; the single failure (`mcp-bridge.test.cjs` "macOS scan…") fails identically on bare `main` — macOS-only test on a non-macOS runner, untouched here. Source-only (no `dist/` xpi), matching your release-time rebuild cadence.

### DIFF
--- a/extension/mcp_server/api.js
+++ b/extension/mcp_server/api.js
@@ -328,8 +328,9 @@ var mcpServer = class extends ExtensionCommon.ExtensionAPI {
                       name: { type: "string", description: "Attachment filename" },
                       contentType: { type: "string", description: "MIME type, e.g. application/pdf" },
                       base64: { type: "string", description: "Base64-encoded file content" },
+                      content: { type: "string", description: "Alias for base64 (accepted for backwards compatibility); base64 takes precedence when both are set" },
                     },
-                    required: ["name", "base64"],
+                    required: ["name"],
                     additionalProperties: false,
                   },
                 ],
@@ -366,8 +367,9 @@ var mcpServer = class extends ExtensionCommon.ExtensionAPI {
                       name: { type: "string", description: "Attachment filename" },
                       contentType: { type: "string", description: "MIME type, e.g. application/pdf" },
                       base64: { type: "string", description: "Base64-encoded file content" },
+                      content: { type: "string", description: "Alias for base64 (accepted for backwards compatibility); base64 takes precedence when both are set" },
                     },
-                    required: ["name", "base64"],
+                    required: ["name"],
                     additionalProperties: false,
                   },
                 ],
@@ -614,8 +616,9 @@ var mcpServer = class extends ExtensionCommon.ExtensionAPI {
                       name: { type: "string", description: "Attachment filename" },
                       contentType: { type: "string", description: "MIME type, e.g. application/pdf" },
                       base64: { type: "string", description: "Base64-encoded file content" },
+                      content: { type: "string", description: "Alias for base64 (accepted for backwards compatibility); base64 takes precedence when both are set" },
                     },
-                    required: ["name", "base64"],
+                    required: ["name"],
                     additionalProperties: false,
                   },
                 ],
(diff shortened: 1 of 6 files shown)


TKasperczyk:
Requesting changes -- reviewed the three-hardening delta now that #125 is in (good news first: this merges clean against current main, no rebase needed, and the full suite passes with it applied). The directions are all right, but each hardening has a real gap:

1. **htmlToMarkdown sanitization is bypassable**: existing backslashes aren't escaped, so link text like `click\](https://attacker.example)` serializes to `[click\\](https://attacker.example)](https://good.example)` -- a CommonMark renderer reads the doubled backslash as an escaped backslash, leaving the `]` active, and the injected link wins. Confirmed against markdown-it. The negative-lookbehind treats *any* preceding backslash as escaping instead of checking odd/even parity. Related: an allowed `https:` URL containing a line break can break out of the `<...>` destination and inject fresh markdown. Escape pre-existing backslashes, reject controls/newlines in destinations, and add renderer-level adversarial tests (markdown-it as a devDependency for tests is fine by me).
2. **The filter forward/reply block doesn't cover all paths**: `updateFilter({enabled: true})` can re-enable an existing forward/reply filter; a conditions-only update copies existing actions verbatim (and can broaden the match scope); `applyFilters` executes the full list without checking for blocked actions; `reorderFilters` can move a blocked action around `stopExecution` rules. On top of that, fields are mutated before replacement actions are validated, so a rejected update still leaves the in-memory filter changed. The pref needs to inspect *existing* actions on update/apply/reorder, and validation has to happen before mutation.
3. **Both new prefs are missing from the experiment schema and options UI**: the getters/setters exist in `api.js`, but `schema.json` jumps straight from `setBlockSkipReview` to `getStableAuthToken`, and `options.html` only exposes `blockSkipReview` -- so the opt-outs your error messages (and `test/poc/README.md`) point at don't exist for normal users. Wire them up.
4. **The F1 PoC leaves the dangerous filter enabled**: `listFilters` returns account wrappers with nested `filters`, the PoC treats it as a flat array, never finds the rule it created, and skips cleanup -- and a `.invalid` recipient doesn't stop Thunderbird from handing the message to the configured SMTP relay if the rule fires. Fix cleanup using the creation result and create the demo filter disabled, or drop the harness from this PR.

Minor, take or leave: `mid:`/`news:`/`sms:`/`sip:`/`xmpp:`/`webcal:` links get flattened to text (fine as a conservative trade-off, but worth a code comment); audit entries record attempts rather than outcomes and `countRecipients` miscounts quoted display names containing commas; the new attachment-summary helper is unused (lint warns); and there are literal control bytes in the URL-cleaning regexes, which makes tools like `rg` treat `api.js` as binary -- use `\x00-\x1f` escapes.

The filter-forward block is the most important fix in this stack (it closes a silent-exfiltration vector), so I want to land it quickly once the coverage gaps are closed. Happy to help if you want to iterate on this!

PR #111, opened 2026-05-17
Enhance release workflow and versioning for builds

the78mole:
This pull request introduces automated build and release workflows, improves documentation for both users and developers, and enhances the build script to support versioning and auto-updates for the Thunderbird MCP extension. The main focus is on making installation, configuration, and release management easier and more robust, especially for end-users and distribution via GitHub Releases.

The mein thing left is the signing of the extension by mozilla, see Feature at the end of this description.

This PR already integrates PR #100 opened 5 days ago by [emercons](https://github.com/TKasperczyk/thunderbird-mcp/issues?q=is%3Apr+is%3Aopen+author%3Aemercons)

**Key changes:**

### CI/CD Automation and Release Management
- Added a GitHub Actions workflow (`.github/workflows/build-and-release.yml`) to automatically build the extension, generate versioned artifacts, and publish releases with semantic versioning and source archives. This streamlines the release process and ensures consistent packaging.
- Enhanced the build script (`scripts/build.sh`) to:
  - Use the `BUILD_VERSION` environment variable for CI builds, ensuring correct version stamping.
  - Dynamically determine the GitHub repository owner and name for generating the add-on ID and update URLs.
  - Automatically update `manifest.json` with version, add-on ID, and update URL, and generate an `updates.json` file for Thunderbird's auto-update mechanism.

### User and Developer Documentation Improvements
- Added a comprehensive usage guide (`.github/skills/extension-usage-guide.md`) detailing setup steps, primary use cases (archiving emails and attachments), advanced search, workflow templates, available tools, configuration, security, and troubleshooting. This guide is designed for both AI assistant integration and end-user onboarding.
- Updated the main `README.md` to:
  - Simplify installation and configuration instructions, including npm-based bridge installation and AI assistant integration.
  - Recommend adding the usage guide as a Copilot skill for natural-language archiving commands.

### Versioning and Packaging Consistency
- Updated the version in `package.json` and `extension/manifest.json` to a placeholder (`99.9.9`) to ensure consistency during the build process. [[1]](diffhunk://#diff-1929f2f40162baf33d4a074cc7a94593fc44ffb991fd7a7f597f90643da7f7f4L4-R4) [[2]](diffhunk://#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519L3-R3)

These changes collectively make the extension easier to install, configure, and keep up-to-date, while also providing clear documentation for both users and contributors.

---

[...]

PR #134, opened 2026-06-10
feat(ci): add build-and-release workflow

the78mole:
Split from #111 as requested: this PR contains only the GitHub Actions build/release workflow.\n\nIncluded:\n- Add .github/workflows/build-and-release.yml to build the extension and publish release assets.\n\nExplicitly excluded from this split:\n- Auto-update wiring (no update_url/updates.json changes)\n- Mozilla signing work\n- package.json -> manifest version path changes\n- Docs and generated artifacts (including dist/thunderbird-mcp.xpi)\n\nThis keeps scope to the pure build workflow so it can be reviewed and merged independently.

### DIFF
--- a/.github/workflows/build-and-release.yml
+++ b/.github/workflows/build-and-release.yml
@@ -1,0 +1,63 @@
+name: Build and Release
+
+on:
+  push:
+    branches:
+      - main
+  workflow_dispatch:
+
+jobs:
+  build-and-release:
+    runs-on: ubuntu-latest
+
+    permissions:
+      contents: write
+
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+
+      - name: Get semantic version
+        id: semver
+        uses: paulhatch/semantic-version@v5.0.3
+        with:
+          tag_prefix: "v"
+          version_format: "${major}.${minor}.${patch}"
+          major_pattern: "/^(feat|fix|refactor)!:/"
+          minor_pattern: "/^feat:/"
+
+      - name: Setup Node.js
+        uses: actions/setup-node@v4
+        with:
+          node-version: "18"
+
+      - name: Build extension
+        run: bash scripts/build.sh
+
+      - name: Rename release files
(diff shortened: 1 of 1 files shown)


TKasperczyk:
Thanks for splitting the build workflow out into this, @the78mole -- it's the basis for where releases are going.

I dug into the release setup properly since. The direction is a **tag-triggered** build+release rather than push-to-main: the workflow fires when a `vX.Y.Z` tag is pushed and builds from the already-bumped `package.json` the tag points at -- which sidesteps the version-skew entirely (no need to thread a computed version into `build.sh`, since the tagged commit already has the right one). Concretely the changes to this PR:

- trigger on `push` tag `v*.*.*` (+ `workflow_dispatch`), with a concurrency gate, instead of push to main (also avoids colliding with the existing `v0.6.0` tag)
- run the tests (`node --test test/*.cjs`) and assert `manifest.version === tag` before building
- drop the hand-rolled source tar/zip -- GitHub auto-attaches source archives to every release
- pin actions to current majors (checkout@v4, setup-node@v4, action-gh-release@v2 by SHA); drop `paulhatch/semantic-version` (versioning handled separately)

Want to take it the rest of the way yourself, or would you rather I push those onto your branch and merge? Either way it lands as your PR -- no rush, just say which.

Aside on #111: your auto-update instinct there was right. Turns out Thunderbird ships `xpinstall.signatures.required=false`, so a self-hosted `update_url` + `updates.json` works with no AMO signing at all (and keeping the current add-on id). I'll follow up on that thread separately.

the78mole:
Sorry, was again fully occupied with other stuff... Just a note on "push to main" trigger... This trigger also fires for main, when you merge a branch back to main.
I usually use paulhatch/semantic-version and if you stick to conventional commit messages, the version bumps and tagging happen automatically. That's just convenient for the maintainer...
In combination with squash-merging, you can decide by your squash-commit-message, what shall be bumped, patch, minor or major... All other stuff is then handled by the workflow... As an example, you can have a look here: https://github.com/the78mole/cpu-loader/tree/main/.github/workflows#versioning-strategy

PR #144, opened 2026-06-16
fix: refresh MCP connection info and support Thunderbird 152

### LINKED ISSUE #143: Bridge becomes unreachable when connection.json disappears while Thunderbird MCP server is still running
## Summary

I've encountered multiple times when I ask Claude to access this MCP and it says it can't find it. This also led to scheduled task failures. Turns out I hit a state on native macOS where Thunderbird was running, the Thunderbird MCP extension HTTP server was still listening on `127.0.0.1:8765`, but the bridge could no longer connect because `<TmpD>/thunderbird-mcp/connection.json` was missing. This is because macOS cleaned up the `connection.json` file.

Restarting Thunderbird immediately recreated `connection.json` and the bridge worked again, so the failure was not the HTTP server or MCP bridge startup. It was the missing auth/port handoff file.

## Observed Behavior

Claude reported the MCP server as unreachable.

Running the bridge directly still handled MCP lifecycle methods:

```json
{"jsonrpc":"2.0","id":1,"result":{"protocolVersion":"2025-06-18","capabilities":{"tools":{}},"serverInfo":{"name":"thunderbird-mcp","version":"0.6.0"}}}
{"jsonrpc":"2.0","id":2,"result":{}}
```

But `tools/list` failed:

```text
Bridge error: Connection discovery failed. Tried: native tmp (.../T/thunderbird-mcp/connection.json): file not found; macOS temp scan (/var/folders/*/*/T/thunderbird-mcp/connection.json): no matching files owned by uid 501. Is Thunderbird running with the MCP extension? The extension must be started first to create the connection file.
```

At the same time, Thunderbird was still listening:

```text
thunderbi ... TCP 127.0.0.1:8765 (LISTEN)
```

And a direct unauthenticated localhost probe returned the expected auth rejection, proving the extension HTTP server was alive:

```json
{"jsonrpc":"2.0","id":null,"error":{"code":-32600,"message":"Invalid or missing auth token"}}
```

After restarting Thunderbird, this file was recreated:

```text
/var/folders/.../T/thunderbird-mcp/connection.json
```

Then `tools/list` and `listAccounts` worked again.

## Expected Behavior

If the Thunderbird MCP HTTP server is still running, the bridge should continue to be able to discover its port/token handoff information.

If the temp `connection.json` file is removed by OS temp cleanup while Thunderbird continues running, the extension should recreate it without requiring a Thunderbird restart.

## Actual Behavior

The server kept running in memory, but the only bridge discovery handoff file disappeared. The bridge correctly failed closed because it had no token, but the system could not self-heal until Thunderbird restarted.

## Environment

- OS: macOS
- Thunderbird: native macOS app
- Thunderbird MCP: `0.6.0`
- Bridge command:

```json
{
  "command": "node",
  "args": ["/Users/sph/Documents/Code/thunderbird-mcp/mcp-bridge.cjs"]
}
```

## Related Issues/PRs Checked

I found related connection discovery issues, but they appear to cover different failure modes:

- #61 / #70: Flatpak paths for `connection.json`
- #65: Snap temp path discovery
- #76: stale candidate masking a live connection file
- #77: stable auth token ergonomics
- #63: older proposal to move connection file out of temp

This issue is specifically about native macOS temp-file disappearance while the extension HTTP server remains alive.

## Possible Fix

Have the extension periodically verify that the live server's `connection.json` still exists and still matches the active `{ port, token, pid }`. If the file is missing, malformed, or stale, rewrite it while preserving the existing fail-closed bridge behavior.

This keeps session-scoped auth intact but prevents OS temp cleanup from stranding a live server.

JGSphaela:
## Summary

- Periodically verifies that the live Thunderbird MCP server's `connection.json` still matches the active `{ port, token, pid }`.
- Rewrites the connection file if it is missing, malformed, or stale while the HTTP server is still running.
- Uses Gecko `nsITimer` for the refresh loop, which works reliably from the experiment parent scope.
- Stops the refresh timer on startup failure, server restart, listen-all rebinding, and extension shutdown.
- Bumps Thunderbird compatibility to `strict_max_version: "152.*"` and rebuilds the bundled XPI.
- Adds regression coverage for the connection-info refresh lifecycle.

## Why

I reproduced a native macOS state where Thunderbird was still running and the MCP extension HTTP server was still listening on `127.0.0.1:8765`, but `<TmpD>/thunderbird-mcp/connection.json` had disappeared.

In that state, the bridge could still answer local lifecycle methods like `initialize` and `ping`, but `tools/list` failed with `Connection discovery failed` because the bridge had no auth/port handoff file to read. A direct unauthenticated probe to `127.0.0.1:8765` returned the expected `403` auth error, confirming the HTTP server was alive.

Restarting Thunderbird recreated `connection.json` and restored the bridge. This patch makes the extension self-heal that missing handoff file without requiring a restart.

While testing Thunderbird 152, I also found that the extension needed the compatibility cap bumped from `151.*` to `152.*`. With the cap bumped and the refresh timer using `nsITimer`, the bridge works on Thunderbird 152 and recovers after `connection.json` is deleted while Thunderbird remains running.

Closes #143

## Implementation

- Adds `CONNECTION_FILE_REFRESH_MS` set to 30 seconds.
- Adds `ensureConnectionInfo(port, token)`, which reads the current connection file and verifies it matches the active server's `port`, `token`, and `Services.appinfo.processID`.
- Reuses the existing secure `writeConnectionInfo()` path when the file is missing, malformed, or stale.
- Tracks a single refresh timer on `globalThis.__tbMcpConnectionInfoRefreshTimer`.
- Uses `Cc["@mozilla.org/timer;1"].createInstance(Ci.nsITimer)` instead of DOM timers in the experiment API parent context.
- Clears the refresh timer anywhere the server is stopped or startup fails.

## Testing

```bash
npm test
```

Latest rebased run while Thunderbird was running:

```text
377 tests
360 pass
17 skipped
0 fail
```

The skipped tests are the existing connection-file mutation tests that skip when Thunderbird is running on the default MCP port. Before the final rebase, the full local suite also passed with Thunderbird stopped:

```text
377 tests
377 pass
0 fail
0 skipped
```

Runtime verification on Thunderbird 152:

- Installed the rebuilt XPI.
- Confirmed Thunderbird listens on `127.0.0.1:8765`.
- Confirmed `mcp-bridge.cjs` `tools/list` returns 39 tools.
- Deleted `<TmpD>/thunderbird-mcp/connection.json` while Thunderbird stayed running.
- Confirmed the extension recreated `connection.json` and `tools/list` still returned 39 tools.


JGSphaela:
Close by #155

TKasperczyk:
Thanks @JGSphaela, and thanks for closing it out. For the record: your fix shipped in `v0.7.2` as a co-authored commit (`0efdefb`). Kept the nsITimer self-heal and lifecycle wiring, moved the freshness decision into a small helper so it gets a real behavioral test, and put the secure-create write back on the initial-start path (the timer self-heals after). Solid repro on #143.

PR #145, opened 2026-06-17
feat(messages): add getMessageHeaders and batchGetMessageHeaders

JordanRO2:
## What

Two read-only, header-only message tools for agent workflows that need to triage or enrich a result set without decoding bodies or paying N round-trips.

- **`getMessageHeaders(messageId, folderPath)`** — returns a flat header object: `id, subject, author, recipients, ccList, date, tags, isRead, isFlagged, threadId, references, inReplyTo, size`. Goes through the same `findMessage → openFolder → getAccessibleFolder` access gate as `getMessage`.
- **`batchGetMessageHeaders(messageIds, folderPath)`** — one folder open + a single linear enumeration fallback (O(M), not O(M·N)) for ids the direct `getMsgHdrForMessageID` lookup misses. Returns `{ headers: { id → header | { error } }, total, failed }`.

Both reuse a shared `msgHdrToHeaderObject` extractor so their field shape matches what the other read tools emit.

## Why

Agents frequently want just the headers (sender / subject / date / threading) for a page of search results — to dedupe, thread, or decide which messages to fully fetch. Today that means N `getMessage` calls, each decoding the body. These tools make header triage a single call.

## Tests

`test/validation.test.cjs` covers both schemas (valid, missing-required, wrong-type, batch shape/shallow-null); `ALL_TOOLS` updated. Full suite: **386 pass, 0 fail**. `eslint .`: **0 errors**.

## Notes

- The 200-id hard cap on `batchGetMessageHeaders` lives in the handler (returns a clear `{ error }`) rather than the schema, so the empty-input fast path (`{ headers: {}, total: 0, failed: 0 }`) stays reachable and over-cap is a normal tool result, not a dispatch rejection.
- `getMessageHeaders` reuses the existing `findMessage`, so its not-found behavior matches `getMessage`.

### DIFF
--- a/extension/mcp_server/api.js
+++ b/extension/mcp_server/api.js
@@ -232,6 +232,38 @@ var mcpServer = class extends ExtensionCommon.ExtensionAPI {
         },
       },
       {
+        name: "getMessageHeaders",
+        group: "messages", crud: "read",
+        title: "Get Message Headers",
+        description: "Read just the header fields (subject, from, to, cc, date, tags, flags, threading) of a single message by its ID. Lighter than getMessage — no body/MIME parse.",
+        inputSchema: {
+          type: "object",
+          properties: {
+            messageId: { type: "string", description: "The message ID (from searchMessages results)" },
+            folderPath: { type: "string", description: "The folder URI path (from searchMessages results)" },
+          },
+          required: ["messageId", "folderPath"],
+        },
+      },
+      {
+        name: "batchGetMessageHeaders",
+        group: "messages", crud: "read",
+        title: "Batch Get Message Headers",
+        description: "Fetch headers for up to 200 messages that share one folder, in a single round-trip. Returns a map keyed by messageId; ids not found in the folder report a per-item error without failing the batch. Pair with searchMessages to enrich a result set without N round-trips.",
+        inputSchema: {
+          type: "object",
+          properties: {
+            messageIds: {
+              type: "array",
+              description: "Message IDs to read headers for (hard cap 200, enforced by the handler). All must live in folderPath.",
+              items: { type: "string", description: "An RFC Message-ID (from searchMessages results)" },
+            },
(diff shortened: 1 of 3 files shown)


TKasperczyk:
Thanks @JordanRO2 -- reviewed now that #125 is in. The tool pair is wanted (#148/#149 build on it, so let's get the base right), but requesting changes on four things:

1. **Rebase needed**: `test/validation.test.cjs` asserts `[null]` array items pass shallow validation -- #125's recursive validator rejects those now, so the branch conflicts with current main.
2. **Unbounded synchronous scan**: when even one ID misses direct lookup, `batchGetMessageHeaders` linearly enumerates up to 50,000 headers on Thunderbird's main thread with no yielding or time budget -- same freeze class as #161 -- and hitting the cap silently reports the remaining IDs as "not found". Make the fallback cooperative (yield every N headers) and truncation-aware (distinguish "not found" from "not scanned"), or drop the deep fallback and document the direct-lookup contract.
3. **Header contract bugs**: `ccList` is returned as the raw DB value -- only subject/author/recipients have RFC 2047-decoded accessors on `nsIMsgDBHdr`, so decode ccList yourself. And `inReplyTo` will essentially always be empty: Thunderbird stores In-Reply-To into `references` rather than an `"in-reply-to"` property, so read the references fallback the way the mail parser populates it.
4. **Shape consistency**: results return `isRead`/`isFlagged`/string-`threadId` where every existing read tool returns `read`/`flagged`/numeric `threadId`. Since #148/#149 reuse `msgHdrToHeaderObject`, this needs settling now -- match the existing shapes.

Also worth doing while you're in there: add `maxItems` to the ids schema (the handler caps at 200, but validation walks every supplied item before the handler sees it), and wrap the handlers so unexpected exceptions surface as the repo's `{error}` form instead of raw JSON-RPC errors.

Happy to review again quickly -- with the shape and scan policy fixed, #148/#149 should each need only one rebase.

Issue #116, opened 2026-05-22
Thunderbird 151.0 reports Thunderbird MCP incompatible because `strict_max_version` is capped at `150.*`


powyncify:
First, thank you for this fantastic add-on! Dziękuję bardzo!

Like #114 , I ran into a compatibility block with Thunderbird MCP on Windows after Thunderbird updated to `151.0`. Here's a diagnosis and a workaround.

### Environment

- Thunderbird: `151.0`
- OS: Windows
- Add-on repo: `TKasperczyk/thunderbird-mcp`
- Add-on ID: `thunderbird-mcp@tkasperczyk.dev`

### Primary issue

Thunderbird reports:

```text
Thunderbird mcp is incompatible with Thunderbird 151.0
```

The manifest currently declares:

```json
"browser_specific_settings": {
  "gecko": {
    "id": "thunderbird-mcp@tkasperczyk.dev",
    "strict_min_version": "102.0",
    "strict_max_version": "150.*"
  }
}
```

That hard cap prevents the add-on from loading on Thunderbird `151.0`, so:

- the add-on never starts
- `connection.json` is never published
- the MCP bridge cannot connect

### Minimal workaround that cleared the version gate

Patch:

```json
"strict_max_version": "151.*"
```

by editing:

```text
extension/manifest.json
```

### Second issue: easy rebuild path produced a corrupt XPI

After patching the manifest, I first rebuilt the `.xpi` using PowerShell `Compress-Archive`. Thunderbird then reported:

```text
This add-on could not be installed because it appears to be corrupt.
```

So there are really two separate problems in this upgrade scenario:

1. The explicit compatibility ceiling blocks Thunderbird 151.0
2. A naive Windows repack path can produce an `.xpi` Thunderbird rejects as corrupt

### Rebuild method that worked

The rebuilt `.xpi` worked when packaged from the **contents of** `extension/` with 7-Zip using ZIP format.

In other words:

- build from `extension/*`
- not from the repo root
- use a real zip archiver such as `zip` or `7z`
- do not rely on PowerShell `Compress-Archive` for this add-on package

Example Windows PowerShell packaging command that worked:

```powershell
$repo = 'C:\path\to\thunderbird-mcp'
$ext = Join-Path $repo 'extension'
$dist = Join-Path $repo 'dist'
$xpi = Join-Path $dist 'thunderbird-mcp.xpi'
$sevenZip = 'C:\Program Files\7-Zip\7z.exe'

[...]

Issue #114, opened 2026-05-21
Not compatible with Thunderbird 151.0


alizehctl:
The addon (installed via the bundled xpi file) was working fine yesterday, but after my Thunderbird got updated, it says this addon is incompatible with Thunderbird 151.0

powyncify:
I noticed the same and just posted a workaround, a temporary fix #116

TKasperczyk:
Resolved by `92cfff3` (PR #112) -- bumped `strict_max_version` to `151.*`. Thanks for the report.

Not in a release yet, rebuild from main and reinstall to verify.

PR #147, opened 2026-06-17
fix(compat): bump strict_max_version to 152.* for Thunderbird 152

### LINKED ISSUE #146: Not compatible with Thunderbird 152.0 (strict_max_version capped at 151.*)
## Summary

After updating to Thunderbird 152.0, the extension reports "Thunderbird MCP is incompatible with Thunderbird 152.0" and will not load.

## Cause

`extension/manifest.json` sets `browser_specific_settings.gecko.strict_max_version` to `151.*`, so Thunderbird 152 falls outside the supported range and the add-on is rejected.

## Environment

- Thunderbird: 152.0
- Extension: v0.7.0

## Fix

Bump `strict_max_version` to `152.*`. This is the same trivial-but-blocking fix as the earlier 150 -> 151 bump (#114, #116).

bitswype:
## Summary

Thunderbird 152 was released and the extension caps compatibility at `strict_max_version: "151.*"`. As a result, TB 152 reports "Thunderbird MCP is incompatible with Thunderbird 152.0" and refuses to load the add-on.

This bumps the cap to `152.*` so users on the new version can install and run the extension. It is the same trivial-but-blocking fix as the earlier 150 -> 151 bump (#114, #116, commit 92cfff3).

Closes #146.

## Change

- `extension/manifest.json`: `strict_max_version` `151.*` -> `152.*`

## Test plan

- Built the XPI from this branch and installed it on Thunderbird 152.0; the add-on now loads instead of being rejected as incompatible.
- Full test suite passes (`npm test`): 375 passing, 0 failing.

### DIFF
--- a/extension/manifest.json
+++ b/extension/manifest.json
@@ -7,7 +7,7 @@
     "gecko": {
       "id": "thunderbird-mcp@tkasperczyk.dev",
       "strict_min_version": "102.0",
-      "strict_max_version": "151.*"
+      "strict_max_version": "152.*"
     }
   },
   "icons": {


TKasperczyk:
Thanks @bitswype -- right fix, and it's shipped in `v0.7.1`. I widened `strict_max_version` to `200.*` rather than `152.*` so we stop hitting this every TB major (the tight cap was only ever there for ATN, and this self-distributes as an unlisted XPI). Also added a weekly CI job that load-tests new Thunderbird releases against the extension, so a real incompatibility gets caught early instead of via a user report. Closing as superseded -- credited in the #146 resolution and the v0.7.1 notes.

PR #148, opened 2026-06-17
feat(messages): add searchByThread and getSenderHistory (stacked on #145)

JordanRO2:
> **Stacked on #145** — please review/merge that first. The `getMessageHeaders` / `batchGetMessageHeaders` commit shown here will drop out of this diff once #145 lands.

## What

Two more read-only, header-only query tools, built on the shared `msgHdrToHeaderObject` extractor introduced in #145.

- **`searchByThread(messageId, folderPath, maxResults)`** — walks the anchor's folder once and returns headers for every message sharing its `threadId` (the anchor is included), newest-first. Returns `{ threadId, anchorId, thread, truncated, scanned }`, or `{ thread: [anchor], threadId: null, anchored: true }` when the anchor has no thread id. Result cap default 100 / max 500; enumeration bounded by a local `SCAN_CAP` of 50000.
- **`getSenderHistory(email, maxResults, scanCap, sinceDays)`** — scans the inbox of each accessible account for messages from a sender (case-insensitive substring on the author field), newest-first. `maxResults` default 50 / max 200; per-account `scanCap` default 5000 / max 50000; optional `sinceDays` window.

Both reuse the existing folder/account access gate, and the caps are enforced imperatively in the handlers (consistent with the rest of the batch surface) rather than via schema bounds.

## Why

These round out header-level triage for agents: "show me the whole thread" and "what has this person sent me recently" without N body fetches.

## Tests

`test/validation.test.cjs` covers both schemas; `ALL_TOOLS` updated (structural test passes). Full suite green, `eslint .` 0 errors.

## Known limitation

`getSenderHistory` stops scanning once `maxResults` matches are collected, so "newest-first across every account" holds within the accounts actually visited rather than globally. Documented in the schema description; happy to make it strictly global if you'd prefer.

## Follow-up

`searchAttachments` (the third tool in this family) is intentionally held back for its own PR — its MIME-parse step needs a bounded-concurrency pass before it's a good upstream citizen.

### DIFF
--- a/extension/mcp_server/api.js
+++ b/extension/mcp_server/api.js
@@ -232,6 +232,38 @@ var mcpServer = class extends ExtensionCommon.ExtensionAPI {
         },
       },
       {
+        name: "getMessageHeaders",
+        group: "messages", crud: "read",
+        title: "Get Message Headers",
+        description: "Read just the header fields (subject, from, to, cc, date, tags, flags, threading) of a single message by its ID. Lighter than getMessage — no body/MIME parse.",
+        inputSchema: {
+          type: "object",
+          properties: {
+            messageId: { type: "string", description: "The message ID (from searchMessages results)" },
+            folderPath: { type: "string", description: "The folder URI path (from searchMessages results)" },
+          },
+          required: ["messageId", "folderPath"],
+        },
+      },
+      {
+        name: "batchGetMessageHeaders",
+        group: "messages", crud: "read",
+        title: "Batch Get Message Headers",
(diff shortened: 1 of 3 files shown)


TKasperczyk:
Holding this until #145's review items land -- the header-shape contract (`read`/`flagged`/numeric `threadId` vs the current `isRead`/`isFlagged`/string) and the cooperative-scan policy both live in `msgHdrToHeaderObject` and the fallback loop this PR builds on, so they need settling in the base first. See the review on #145.

After #145 merges, rebase and this should shrink to its own ~290 lines. Heads-up for that rebase: main's validation layer changed today (#125 plus follow-ups).

PR #149, opened 2026-06-17
feat(messages): add searchAttachments with bounded MIME-parse concurrency (stacked on #145)

JordanRO2:
> **Stacked on #145** — please review/merge that first; the `getMessageHeaders` / `batchGetMessageHeaders` commit shown here drops out of this diff once #145 lands. (Independent of #148.)

## What

**`searchAttachments(nameContains, contentType, folderPath, maxResults, scanCap)`** — find messages whose attachments match a filename substring and/or a Content-Type prefix, scanning a folder (or the inbox of each accessible account when `folderPath` is omitted). Returns header objects (via the shared `msgHdrToHeaderObject` from #145) annotated with the matching attachment metadata, newest-first.

Caps: `maxResults` default 50 / max 200; `scanCap` default 5000 / max 50000 **per folder**; candidates pre-filtered on the Attachment flag with a `cap*4` buffer; results sliced to `cap`.

## Bounded MIME-parse concurrency (the interesting part)

Resolving attachment metadata requires a `MsgHdrToMimeMessage` parse per candidate. A naive implementation fires one parse per candidate at once — up to `cap*4` (≈800) **simultaneous** parses, which on IMAP means a burst of network fetches. This PR bounds that:

- a **worker pool of `MAX_MIME_CONCURRENCY = 8`** pulls candidates from a shared index, so at most 8 parses are ever in flight;
- each parse is guarded by a **one-shot `nsITimer` timeout** so a stuck parse frees its slot instead of wedging the batch;
- an **aggregate deadline** stops pulling new work so the call can never exceed the client/bridge request timeout (`truncated` is set when it trips);
- the **3-arg `MsgHdrToMimeMessage` form** is used (no `allowDownload`), so it reads cached MIME without forcing network body downloads.

## Tests

`test/validation.test.cjs` covers the schema; `ALL_TOOLS` updated (structural test passes). Full suite green, `eslint .` 0 errors.

> Note: the bounded concurrency + timeouts are a deliberate hardening over a straight port. The concurrency/termination logic was reviewed adversarially (worker-pool accounting, timer/callback race, no early-finish/hang/lost-candidate). Happy to tune `MAX_MIME_CONCURRENCY` / timeout / deadline to your preference.

### DIFF
--- a/extension/mcp_server/api.js
+++ b/extension/mcp_server/api.js
@@ -232,6 +232,38 @@ var mcpServer = class extends ExtensionCommon.ExtensionAPI {
         },
       },
       {
+        name: "getMessageHeaders",
+        group: "messages", crud: "read",
+        title: "Get Message Headers",
+        description: "Read just the header fields (subject, from, to, cc, date, tags, flags, threading) of a single message by its ID. Lighter than getMessage — no body/MIME parse.",
+        inputSchema: {
+          type: "object",
+          properties: {
+            messageId: { type: "string", description: "The message ID (from searchMessages results)" },
+            folderPath: { type: "string", description: "The folder URI path (from searchMessages results)" },
+          },
+          required: ["messageId", "folderPath"],
+        },
+      },
+      {
+        name: "batchGetMessageHeaders",
+        group: "messages", crud: "read",
+        title: "Batch Get Message Headers",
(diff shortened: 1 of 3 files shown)


TKasperczyk:
Same coordination note as #148: waiting on #145's review items first, since this reuses `msgHdrToHeaderObject` and the header-shape contract needs settling in the base before more code builds on it.

The bounded-concurrency MIME parse is the part of this stack I'm most interested in -- it's the right instinct for the same large-folder problem #161 attacks from the search side. Rebase once #145 lands (main's validation layer changed today: #125 plus follow-ups).

PR #150, opened 2026-06-17
feat(messages): add refreshFolder tool

JordanRO2:
## What

**`refreshFolder(folderPath, timeoutMs)`** — force an IMAP fetch on a folder so the server's current state syncs into Thunderbird's local cache. Handy before `searchMessages` / `getMessage` on an account whose offline copy may be stale.

- Drives the fetch with `folder.getNewMessages(null, urlListener)` and detects completion via an `nsIUrlListener` (`OnStopRunningUrl`).
- Bounded by a one-shot `nsITimer`: `timeoutMs` default 15000, **hard-capped at 25000** to stay under the stdio bridge's 30s request timeout — so a slow server yields a structured `{ error: "refreshFolder timed out…" }` instead of the bridge hard-failing. A resolve-once guard cancels the timer on completion.
- Non-IMAP folders short-circuit to `{ success: true, skipped: "not an IMAP folder" }`.

Returns `{ success, folderPath, totalBefore, totalAfter, newMessages }` on a completed fetch, or a structured `{ error, folderPath }` on timeout / fetch-failure / access-denied. Reuses `getAccessibleFolder` for the per-account access gate.

## Tests

`test/validation.test.cjs` covers the schema; `ALL_TOOLS` updated (structural test passes). Full suite green, `eslint .` 0 errors.

Independent of my other open tool PRs (#145/#148/#149) — branches off `main` directly.

### DIFF
--- a/extension/mcp_server/api.js
+++ b/extension/mcp_server/api.js
@@ -666,6 +666,20 @@ var mcpServer = class extends ExtensionCommon.ExtensionAPI {
         },
       },
       {
+        name: "refreshFolder",
+        group: "messages", crud: "read",
+        title: "Refresh Folder",
+        description: "Fetch new mail for an IMAP folder from the server and wait for completion. Returns counts of messages before/after and how many new messages arrived. Non-IMAP folders (POP/Local) are skipped. The wait is bounded by timeoutMs (default 15000, hard-capped at 25000); on timeout the in-flight fetch keeps running in Thunderbird and a timeout result is returned.",
+        inputSchema: {
+          type: "object",
+          properties: {
+            folderPath: { type: "string", description: "Folder URI (from listFolders) to refresh" },
+            timeoutMs: { type: "integer", description: "Max ms to wait for the fetch to complete (default 15000). Hard-capped at 25000 to stay below the stdio bridge's HTTP request timeout (30000ms); larger values are clamped." },
+          },
+          required: ["folderPath"],
+        },
+      },
+      {
         name: "createFolder",
         group: "folders", crud: "create",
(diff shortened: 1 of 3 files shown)


TKasperczyk:
Requesting changes -- self-contained and nicely error-shaped, but the core call is the wrong Thunderbird operation:

1. **`folder.getNewMessages(...)` is the account-level "Get Mail" path**, not a folder refresh: it locates and updates the *Inbox* (then optionally traverses check-new folders), so for a non-Inbox target the listener is attached to a different folder than the work being launched and can simply time out. The targeted, waitable API is `nsIMsgImapMailFolder.updateFolderWithListener`. Note also that main already has `refreshImapFolderSync` calling the target folder's `updateFolder(null)` -- build on or replace that rather than adding a parallel refresh path.
2. **Rebase needed**: the `test/validation.test.cjs` changes re-add integer validation that #125 now does recursively -- conflicts with current main.
3. **`newMessages` is a net folder-count delta**, not arrivals -- an expunge or a filter moving mail can zero it out even when new messages landed. Rename it (`netMessageDelta`) or derive a real new-message signal.
4. **The timeout bounds the promise, not the work**: the fetch keeps running after timeout, and repeated calls have no in-flight guard. With the account-level call from (1) that's a real foot-gun on huge mailstores; with `updateFolderWithListener` it gets much cheaper, but an in-flight guard is still worth having.

Smaller: reject IMAP server roots / no-select folders explicitly, and expose `timeoutMs` bounds in the schema.

The tool itself is a good idea (pairs naturally with #161's cooperative search). Happy to review again quickly!

PR #151, opened 2026-06-17
feat(messages): add dryRunCompose (read-only compose preflight, stacked on #125)

JordanRO2:
> **Stacked on #125** — it reuses the deny-list / recipient-count helpers from that PR (`isSensitiveFilePath`, `countRecipients`, `MAX_FILE_PATH_ATTACHMENT_BYTES`). Please review/merge #125 first; its commits drop out of this diff once it lands.

## What

**`dryRunCompose(to, subject, body, cc, bcc, isHtml, from, attachments)`** — validate compose parameters **without any side effect**: it does not send, save a draft, open a compose window, write temp files, or copy/decode attachment bytes. A pure read-only preflight for agents that want to know "would this send cleanly?" before committing.

Returns a structured report:
- `wouldSucceed` (roll-up), `blockers[]` (identity/resolution failures)
- `resolvedIdentity` (resolved `from`, or the default-identity preview)
- `subjectAfterSanitization`, `bodyLength` (content is **not** echoed), `recipients: {to, cc, bcc}` counts
- per-attachment verdict `{ kind, name, status: ok|blocked|missing|error, reason, size }` — file paths are checked against the sensitive-path deny-list and the 50 MB cap via a **stat-only** `nsIFile` (never opened/read/normalized); inline entries' decoded size is **estimated arithmetically** and checked against the 25 MB cap (base64 is never decoded)
- `skipReviewBlocked` pref snapshot (reported only)

## No side effects (verified)

Deliberately does **not** call the `filePathsToAttachDescs` path (which writes temp files / builds file URIs). The attachment loop only stats paths and arithmetic-estimates inline sizes. Reviewed adversarially specifically to disprove any send/save/window/temp-file/network effect.

## Tests

`test/validation.test.cjs` covers the schema (incl. the content-alias on inline attachments, consistent with the other compose tools); `ALL_TOOLS` updated. Full suite green, `eslint .` 0 errors.

### DIFF
--- a/extension/mcp_server/api.js
+++ b/extension/mcp_server/api.js
@@ -328,8 +328,9 @@ var mcpServer = class extends ExtensionCommon.ExtensionAPI {
                       name: { type: "string", description: "Attachment filename" },
                       contentType: { type: "string", description: "MIME type, e.g. application/pdf" },
                       base64: { type: "string", description: "Base64-encoded file content" },
+                      content: { type: "string", description: "Alias for base64 (accepted for backwards compatibility); base64 takes precedence when both are set" },
                     },
-                    required: ["name", "base64"],
+                    required: ["name"],
                     additionalProperties: false,
                   },
                 ],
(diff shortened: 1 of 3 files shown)


TKasperczyk:
#125 just merged, so this can finally shed the stacked commits -- could you rebase onto current main? The diff should collapse to just the dryRunCompose bits, which makes it reviewable on its own.

I want this one -- a read-only preflight that validates and echoes what *would* be sent fits naturally with the `blockSkipReview` flow. One thing to double-check while rebasing: the validation layer moved under you today (#125's `validateAgainstSchema` plus follow-up hardenings landing on main), so re-run your suite on top of current main before pushing.

PR #153, opened 2026-06-24
ci: automated Thunderbird version tracking and compatibility testing

Oaklight:
## Summary

Adds two GitHub Actions workflows to prevent the extension from silently breaking when Thunderbird releases a new major version (as happened with TB 152 vs `strict_max_version: 151.*`).

### `tb-version-check.yml` — Detect → Test → PR/Issue pipeline

Runs weekly (Monday 08:23 UTC) and on-demand. Three stages chained via `needs:`:

```
detect:  poll Mozilla API, compare latest TB vs strict_max_version
   ↓ (only if bump needed)
compat-test:  bump manifest in-memory, build xpi,
              load into real TB stable+beta via xvfb,
              verify extension is not appDisabled
   ↓
open-pr:   (if both channels pass) auto-create PR with bump
open-issue: (if any channel fails) file a bug with diagnostics
```

Key design: the PR is never opened unless the bumped extension actually loads in both stable and beta Thunderbird. If it fails, an issue is filed instead, pointing to the specific workflow run for debugging.

### `tb-compat-test.yml` — Standalone smoke test for PRs

Triggered on PRs that touch `extension/**`, `package.json`, or `scripts/build.sh`. Same test logic:

- Install Thunderbird stable + beta in CI (Mozilla APT repo / tarball)
- Build XPI, install into temp profile with `extensions.experiments.enabled = true`
- Launch TB headless via xvfb, wait for `extensions.json`
- Assert `appDisabled: false` and `active: true`

### Motivation

TB 152 was released while `strict_max_version` was `151.*`, causing `appDisabled: true` for all users who updated. The extension silently stopped working with no user-facing error. These workflows catch that proactively.

## Test plan

- [ ] Trigger `tb-version-check` manually via `workflow_dispatch` — should detect TB 152 > 151, run compat-test, then open PR
- [ ] Trigger `tb-compat-test` manually — should install TB stable + beta, load extension, pass
- [ ] Verify that if compat-test fails, an issue is created instead of a PR


TKasperczyk:
Shipped a reworked version in `v0.7.2` -- thanks @Oaklight. Landed as a co-authored commit (`8de6987`). Since the manifest cap is now broad (`200.*`, self-distributed), I dropped the auto-bump/auto-PR machinery and kept the part that matters: the real-Thunderbird load test. It now asserts the extension actually activates (`active:true`, not just `appDisabled:false`); the weekly job watches stable/beta/ESR and files a deduplicated issue only on real breakage; the PR smoke test is `pull_request` + `contents:read` (no pwn-request). It already earned its keep -- it confirmed the extension still loads on TB 153.0b3, so the experiment-API removal isn't live in beta yet. Closing as superseded.

PR #154, opened 2026-06-25
fix(httpd): accept port-mapped Host headers in listenAll mode

ticapix:
Hello,

This is an awesome extension !

I'm using https://github.com/jlesage/docker-thunderbird/tree/master to sync my emails on disk (gmail, O365, exchange, protonmail)
My docker compose is
```yaml
services:
  thunderbird:
    image: jlesage/thunderbird:v26.03.1
    restart: unless-stopped
    ports:
      - 7290:5800 # VNC
      - 7293:8765 # MCP
    volumes:
      - config:/config:rw
```

So the MCP server is mapped on the host on the port 7293.

When started with startAll() / listenAll, the server binds to 0.0.0.0 to support Docker, WSL, and reverse-proxy access. Clients reaching it through a port-mapping layer send a Host header whose port reflects the external mapping, not the internal socket port. The identity.getScheme() lookup found no match for that external port and threw HTTP 400, making listenAll non-functional for any port-mapped setup.

Fix: when identity._host is "0.0.0.0" and the Host port is unrecognised, fall back to the primary identity instead of rejecting. The auth token (Authorization: Bearer) is the real security boundary; virtual-host routing protection is not relevant with a single handler.

The loopback-only path (start / start_ipv6) is unchanged.


TKasperczyk:
Shipped in `v0.7.2` -- thanks @ticapix. Landed as a co-authored commit (`e5f8b1c`) rather than a branch merge: kept your 0.0.0.0 -> primary-identity fallback, pulled the host/identity resolution into a helper so it gets a real test, and added a comment making explicit that the Host header isn't a security boundary in listenAll mode (the Bearer token is). Loopback stays strict, malformed hosts still 400. Closing as superseded by the release -- credit's on the commit.

ticapix:
That's perfect :) and thank you for the credit 👍

PR #160, opened 2026-07-03
feat: bridge-local lifecycle tools (launch/stop/status)

Oaklight:
## Summary

Adds three MCP tools handled entirely within the bridge for managing the Thunderbird process:

- **`launchThunderbird`** — starts Thunderbird if not running, waits up to ~30s for the extension to become reachable. Spawns a detached process so Thunderbird survives bridge exit (user may have unsaved work).
- **`stopThunderbird`** — gracefully stops Thunderbird. Tool description explicitly requires user confirmation before calling.
- **`thunderbirdStatus`** — returns `{ running, extensionReachable, port? }`.

When Thunderbird is not running, `tools/list` returns only the 3 lifecycle tools so the AI client can discover and call `launchThunderbird`. Once Thunderbird starts, the extension's 35+ tools are merged in alongside the lifecycle tools.

### Why

Currently the bridge fails after 5 retries if Thunderbird isn't running. Users need wrapper scripts or manual launch. With lifecycle tools, any MCP client can start/stop Thunderbird on demand without special config.

### Design decisions

- **Tools over CLI flags**: on-demand control is more flexible than `--auto-launch`, and the AI can also stop Thunderbird when done
- **Detached spawn**: Thunderbird outlives the bridge — closing Claude/Hermes doesn't kill the user's mail client
- **Auth errors propagated**: 403 from wrong tokens are not swallowed by the `tools/list` fallback

## Test plan

- [x] `npm test` — 398 tests pass (381 pass, 17 pre-existing skips, 0 fail)
- [x] `npx eslint mcp-bridge.cjs test/lifecycle-tools.test.cjs` — no errors
- [x] Manual: kill Thunderbird → bridge returns 3 lifecycle tools only → `launchThunderbird` starts it → full tool list available
- [x] Manual: `stopThunderbird` with user confirmation prompt

### DIFF
--- a/README.md
+++ b/README.md
@@ -115,6 +115,29 @@ Account and tool access are configured via the extension settings page (Tools > 
 
 The same settings page has a "Send Safety" section: toggle **Block `skipReview`** to reject `sendMail` / `replyToMessage` / `forwardMessage` calls that pass `skipReview: true`. The review-window path still works, so clients stay usable -- they just can't send silently.
 
+### Lifecycle management
+
+The bridge exposes three tools for managing the Thunderbird process itself. These are handled locally by the bridge and never forwarded to the extension, so they work even when Thunderbird is not running.
+
+| Tool | Description |
+|------|-------------|
+| `launchThunderbird` | Start Thunderbird if not already running. Waits up to ~30 s for the MCP extension to become reachable. No-op if already running. |
+| `thunderbirdStatus` | Check whether Thunderbird is running and the extension is reachable. Returns `{ running, extensionReachable, port? }`. |
+| `stopThunderbird` | Gracefully stop Thunderbird. **Always requires explicit user confirmation** -- the user may have unsaved work. |
+
+Typical AI workflow:
+
+```
+thunderbirdStatus → not running →
+launchThunderbird → Thunderbird starts, extension connects →
+... use email/calendar/contacts tools ... →
+stopThunderbird (with user confirmation)
+```
+
+Thunderbird is launched as a detached process -- it stays running even if the MCP client (Claude, Hermes, etc.) disconnects. This is intentional: the user may have Thunderbird open for their own work.
+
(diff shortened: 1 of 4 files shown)


Oaklight:
Hey, I've only been able to test this on Linux (Arch, where the process name is `thunderbird-bin`). The platform-specific paths for macOS and Windows are based on docs and conventions but haven't been verified on real hardware:

- **macOS**: uses `open -a Thunderbird` to launch, `osascript -e 'tell application "Thunderbird" to quit'` for graceful stop. Not sure if `pgrep -x thunderbird` picks up the right process name from the app bundle.
- **Windows**: uses `start thunderbird.exe` (relies on App Paths registry) to launch, `taskkill /IM thunderbird.exe` (graceful first, then `/F`) to stop.

Would appreciate if you (or anyone with access to macOS / Windows) could give the lifecycle tools a quick spin and flag anything that doesn't work. Happy to fix whatever comes up.

TKasperczyk:
Thanks @Oaklight -- the DI'd process layer and the degraded `tools/list` when Thunderbird is down are both well thought out, and always requiring confirmation on `stopThunderbird` is the right instinct.

Two things before this can merge:

1. **Opt-in gate**: process lifecycle is a real trust-surface step for the bridge -- today it only proxies HTTP; with this it spawns and kills OS processes. I want the capability, but behind an explicit opt-in (an env var like `THUNDERBIRD_MCP_LIFECYCLE=1` or a bridge flag), with the three tools absent from `tools/list` unless enabled. Default-off keeps the surprise factor at zero for existing setups.
2. **Platform verification**: you flagged the macOS/Windows paths as doc-based guesses -- honest, appreciated. `open -a`/`tasklist`/`taskkill` differ enough in practice (and Flatpak/Snap installs on Linux may not match a plain `pgrep thunderbird` either) that I'd want either one verified run per platform, or the unverified platforms explicitly returning an "unsupported platform" error instead of best-effort guessing.

Happy to review quickly once the opt-in gate is in.

PR #162, opened 2026-07-09
feat(connection): support remote host entries in connection.json

mooons:
Adds support for remote hosts in connection.json by updating `mcp-bridge.cjs` to honor an optional host field in `connection.json` instead of always dialing `127.0.0.1`. The bridge now validates and normalizes host values, while preserving the existing localhost fallback when no host is specified.

Codex wrote the code and tests.

Tested that it works for me over LAN. :)

### DIFF
--- a/mcp-bridge.cjs
+++ b/mcp-bridge.cjs
@@ -10,8 +10,9 @@ const http = require('http');
 const fs = require('fs');
 const path = require('path');
 const os = require('os');
+const net = require('net');
 
-const THUNDERBIRD_HOSTS = ['127.0.0.1'];
+const DEFAULT_THUNDERBIRD_HOSTS = ['127.0.0.1'];
 const REQUEST_TIMEOUT = 30000;
 const CONNECTION_RETRY_DELAY_MS = 1000;
 const CONNECTION_MAX_RETRIES = 5;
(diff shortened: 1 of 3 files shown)


TKasperczyk:
Thanks @mooons -- this is the better-engineered half of the remote-host story: real hostname/IP validation, IPv6 bracket normalization, and actual tests (the `{ concurrency: false }` hardening of the existing suites is appreciated too).

Where I want the feature to land overall: #163 solves the same Docker problem with a bridge-side env var, and the two compose naturally as layers --

1. `THUNDERBIRD_MCP_HOST` env override (whoever runs the bridge container knows the topology),
2. else a validated `host` from `connection.json` (your piece -- nothing writes it today, but it makes the file a complete connection descriptor),
3. else `127.0.0.1`.

Would you be up for folding the env-var layer into this PR, with `normalizeConnectionHost` applied to it as well? An env var deserves the same validation scrutiny as the file. That makes this the single canonical remote-host PR, and I'll credit #163 in the merge.

FYI `extensions.thunderbird-mcp.listenAll` already exists on the extension side for binding beyond loopback, so the pair (listenAll + your host override) is the complete recipe -- worth a short README section in this PR. CI is approved and running now.

PR #163, opened 2026-07-09
External thunderbird host

LaurensdeGraaff:
I had a issue where I have thunderbird running in an other docker container but the MCP can't connect.
This MCP server doesn't include a full on thunderbird app, correct?

By changing this. The MCP can communicate to the docker host by setting the docker service name as `HOST` instead of the local IP.

This is now working for me, feel free to comment. 
Per default, nothing is changed. Only when the user sets the specific `THUNDERBIRD_MCP_HOST` value.
Tried my best to also document this.
I did use AI for this since English is not my native tong.

I have this version running stable for the last 5 days now.

### DIFF
--- a/README.md
+++ b/README.md
@@ -144,6 +144,27 @@ Add to your MCP client config (e.g. `~/.claude.json` for Claude Code):
 }
 ```
 
+If the bridge runs outside Thunderbird's network namespace (for example, in a
+separate Docker container), set `THUNDERBIRD_MCP_HOST` so bridge HTTP requests
+target Thunderbird correctly instead of the bridge container's own loopback:
+
+```json
+{
+  "mcpServers": {
+    "thunderbird-mail": {
+      "command": "node",
+      "args": ["/absolute/path/to/thunderbird-mcp/mcp-bridge.cjs"],
+      "env": {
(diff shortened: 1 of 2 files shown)


TKasperczyk:
Thanks -- the env var is the right control for the Docker case (the bridge operator is the one who knows the topology), and this is exactly the use case I want covered.

Coordination note: #162 attacks the same problem from the `connection.json` side and already has host validation + tests in place, so I've asked over there to fold a `THUNDERBIRD_MCP_HOST` env override -- your layer, credited -- on top of that plumbing. Precedence: env var -> `connection.json` `host` -> `127.0.0.1`.

Keeping this open until that lands so the use case doesn't fall through the cracks; if #162 goes quiet I'll take this PR and add the validation to it instead. FYI `extensions.thunderbird-mcp.listenAll` on the extension side is the other half of the recipe (binding beyond loopback) -- worth pairing with whatever documents this. CI approved and running.

LaurensdeGraaff:
How is it going? What is the status of this?
If you want it programmed differently let me know what changes you expect

harrisonhjones:
Seems like we would also want a way to specify the PORT and SESSION_TOKEN via env if we are going to support the mcp server running on another host (where the session file isn't available?) Looking forward to this feature.

PR #164, opened 2026-07-11
[AI-authored] Add saveMessage bridge tool: write .eml + attachments to a chosen directory

Gunther-Schulz:
> ⚠️ **AI-authored PR** — This pull request was written by an AI coding assistant
> (Claude, via Claude Code) on behalf of @Gunther-Schulz. Please review with that
> in mind; happy to adjust anything.

## Summary
Adds a bridge-local `saveMessage` tool that writes a message's full `.eml` and/or
its attachments to a caller-chosen directory on the machine running the bridge.

## Motivation
`getMessage(saveAttachments: true)` can only write attachments to the OS temp dir,
and `rawSource` returns the message as text — there's no way to save an `.eml` (or
attachments) straight to a target folder. Archiving mail into project folders
otherwise means round-tripping base64 through the MCP client. This closes that gap.

## Why the bridge (not the extension)
- The bridge is the local-machine component and already reads/writes host files
  (`inlineAttachmentPaths` / `ATTACHMENT_TOOLS`), so a user-chosen local path fits here.
- Plain Node `fs` → unit-testable under `node --test` (the extension's XPCOM file
  APIs can't run in CI).
- No `.xpi` change: the tool is appended to `tools/list` and handled before the
  forward, reusing the extension's existing `getMessage` under the hood.

## Tool
`saveMessage({ messageId, folderPath, destDir, saveEml=true, saveAttachments=false, emlFilename?, overwrite=false })`
— writes the raw `.eml` (name from `emlFilename`, else decoded Subject, else
`messageId`) and/or copies attachments into `destDir` (created if missing).
Filenames are sanitized to one path segment; `overwrite:false` refuses to clobber
(`wx`); attachment collisions get ` (2)`/` (3)` suffixes. Returns
`{ emlPath, attachments, skippedAttachments, destDir }`.

## Tests
New `test/save-message.test.cjs` (11 tests): exact rawSource write, Subject-derived
+ explicit filenames, header unfolding, nested `destDir` creation, overwrite
semantics, attachment copy + collision suffix, `{error}` propagation,
missing-rawSource error, arg validation. Full suite: **382 pass / 0 fail**.

## Notes
`destDir` is intentionally an arbitrary local path (mirrors `saveAttachments`
writing to temp). No new dependencies.

---


TKasperczyk:
Thanks @Gunther-Schulz -- this fills a real gap (temp-dir-only `saveAttachments` is limiting), and the implementation reuses the right plumbing (`forwardToThunderbird` + the existing `getMessage` rawSource/saveAttachments paths) with real tests against the actual `saveMessageToDisk` code. Requesting changes though -- a security review turned up blockers, mostly around the fact that as-is this hands an LLM-driven client an unconditional filesystem-write primitive:

1. **Needs a default-off capability gate**: `destDir` is only checked for being a non-empty string, the tool is always advertised, and there's no way to disable it -- an injected tool call can write attacker-supplied attachment bytes to `~/.bashrc`, `~/.ssh/`, autostart entries, anywhere the user can write. Same direction I asked for on #160: opt-in env var (e.g. `THUNDERBIRD_MCP_SAVE_MESSAGE=1`), tool absent from `tools/list` unless enabled. A configurable writable root (`THUNDERBIRD_MCP_SAVE_ROOT`) that `destDir` must resolve inside would be stronger still -- I'd take both.
2. **Raw `.eml` bytes get corrupted**: the extension returns `rawSource` as a Latin-1 byte string, and writing it as UTF-8 expands every byte above `0x7f` (verified: raw tail `80 ff` became `c2 80 c3 bf` on disk). Write `Buffer.from(rawSource, "latin1")` and add a binary-byte regression test -- as-is, 8-bit bodies, MIME parts, and signatures break.
3. **Attachment collision handling is race/symlink-unsafe**: the `.eml` path uses atomic `wx` (good), but attachments do `existsSync` + `copyFileSync`, which follows and overwrites symlinks despite `overwrite: false` (verified with probes, including a symlink introduced between the two calls). Use `COPYFILE_EXCL` and retry the suffix on `EEXIST`; define no-follow semantics for overwrite mode; resolve `destDir` symlinks before any policy check. Also `overwrite: true` disables `usedNames` entirely, so two same-named attachments clobber each other -- and can clobber the `.eml` written earlier in the same call.
4. **Permissions**: new dirs/files land `0755`/`0644` under a normal umask -- saved email shouldn't be world-readable on multi-user systems. `0700`/`0600`, matching the extension's attachment handling.
5. **Windows-safe naming**: the sanitizer contains traversal well (verified: separators, controls, UNC, drive paths all neutralized), but misses `<>:"|?*`, reserved device stems (`CON`, `COM1`, ... including with extensions), trailing dots, and NTFS alternate-stream colons. Also `data.subject` arrives already RFC 2047-decoded from Thunderbird, so the bridge's own partial decoder can go.
6. **Small stuff**: the single CI lint error (missing `{ cause }` on the rethrown `EEXIST` error at `mcp-bridge.cjs:824`), and some handler-level tests (nothing currently exercises the JSON-RPC handler, the `tools/list` injection, or the error envelope).

The core design -- bridge-local, DI'd, `.eml` + attachments split, collision suffixes -- is right, so with the gate, the latin1 fix, and safe collision handling this can merge quickly. Happy to help if you want to iterate on this!

Gunther-Schulz:
Superseded by #184 — all six blockers addressed. Opened as a new PR rather than force-pushing this branch, so the review above still lines up with the code it was written against; #184's first commit is this PR unchanged and the second is the response to your review.

PR #170, opened 2026-07-17
fix(windows): skip POSIX tmp dir permission check on WINNT

b3n0:
## Summary
- Skip the POSIX group/world permission hardening for the thunderbird-mcp tmp directory on Windows (WINNT).
- On Windows, 
sIFile.permissions returns synthetic bits rather than real POSIX modes, so the check falsely refused to write connection.json and broke Server Status.
- The hardening remains active on Linux/macOS where it is meaningful.

## Test plan
- [x] Built extension locally (
ode scripts/build-xpi.cjs)
- [x] Verified patched code is present in dist/thunderbird-mcp.xpi
- [ ] Reinstall extension on Windows and confirm Server Status shows a port without the tmp permissions error
- [ ] Confirm Linux/macOS behavior is unchanged


Free-Max367:
Independent confirmation on a newer Thunderbird than the ones reported so far.

Environment: Windows 11 Pro 25H2 (26200.9168), Thunderbird 154.0 (64-bit).

I applied this exact guard to a rebuilt 0.7.4 XPI.
The server binds and writes `connection.json` on every start, including with the pre-existing `%TEMP%\thunderbird-mcp` directory that triggers the bug. The same build without the guard failed at startup with `Port: --`.

That covers the Windows reinstall item still unchecked in your test plan.
The MCP client also sees the full tool list end to end.

PR #176, opened 2026-07-25
Fix Windows: skip POSIX tmp-dir permission check that blocks server start

Joeclinton1:
## Problem

On Windows the MCP server fails to start on every launch after the first. The extension is active but nothing binds the port and no `connection.json` is written. The error console shows:

```
```

## Cause

`writeConnectionInfo()` refuses to write when the existing tmp directory has group/world bits (`mode & 0o077`). On Windows, `nsIFile.permissions` returns a synthesized POSIX mode that always reports `0o777` for a writable directory and can't be narrowed, so the check is a permanent false positive. The first start works (directory created fresh, check skipped); every start afterward throws, and `start()`'s catch stops the server and deletes the connection file.

## Fix

Guard the POSIX hardening block with `Services.appinfo.OS !== "WINNT"` so it runs only where `nsIFile` exposes real POSIX modes. Linux/macOS behavior unchanged.

## Testing

Thunderbird 140.12.1 (MSIX) on Windows 11: server now binds `127.0.0.1:8765` and writes `connection.json` on every start, including when the tmp directory already exists.


Joeclinton1:
Duplicate of #170, sorry for opening

Issue #181, opened 2026-07-28
v0.7.4: "tmp directory has group/world permissions" error on Windows — POSIX check always fails, connection.json never written


Ctirad-Prochazka:
## Environment
- Thunderbird 153.0 (64-bit), Windows 11 (26200.x)
- thunderbird-mcp v0.7.4 (build 41c73ae)

## Symptom
After updating to v0.7.4, the options page shows:

> Error: thunderbird-mcp tmp directory has group/world permissions — refusing to write connection info

Server Status shows "Running" but Port stays `--` and `connection.json` is never written, so the bridge cannot connect. Fixing NTFS ACLs on `%LOCALAPPDATA%\Temp\thunderbird-mcp` (e.g. `icacls ... /inheritance:r /grant:r "<user>:(OI)(CI)F"`) has no effect.

## Root cause
In `writeConnectionInfo()` (extension/mcp_server/api.js), the POSIX hardening added in v0.7.4 assumes:

```javascript
// permissions is 0 on platforms that don't expose POSIX modes
// (Windows ACLs), so the chmod is a no-op there.
const mode = tmpDir.permissions;
if (mode && (mode & 0o077) !== 0) { ... }
```

However, on Windows `nsIFile.permissions` is not 0 — it's emulated from the read-only file attribute. A writable directory always reports `0777` regardless of the actual NTFS ACLs, and setting `tmpDir.permissions = 0o700` cannot clear the emulated group/world bits. As a result, the check **always throws on Windows whenever the directory already exists** (i.e. on every startup after the very first one).

## Suggested fix
Skip the POSIX mode check on Windows, where security is governed by NTFS ACLs:

```javascript
const isWindows = Services.appinfo.OS === "WINNT";
const mode = tmpDir.permissions;
if (!isWindows && mode && (mode & 0o077) !== 0) {
```

I've verified this patch locally (rebuilt XPI as 0.7.4.1): the error disappears, `connection.json` is written, and the port is registered correctly.

Optionally, on Windows the directory ACLs could be tightened via a different mechanism, but the current POSIX-mode approach cannot work there.

Po odeslání si poznamenejte číslo issue – až vyjde oficiální 0.7.5 s opravou, auto-update vaši lokální 0.7.4.1 přepíše a vše se srovná.

[...]

Issue #178, opened 2026-07-26
v0.7.4 regression: MCP server cannot start on Windows — "tmp directory has group/world permissions"


Martin-Koch-Roundit:
Summary

On Windows, v0.7.4 fails to start the MCP server. `writeConnectionInfo()` throws,
no `connection.json` is written, and every MCP client sees an empty tool list.

This is a regression: v0.7.3 works on the same machine.

## Environment

- Windows 11
- Thunderbird `153.0`
- thunderbird-mcp v0.7.4 (installed from `dist/thunderbird-mcp.xpi`)
- Client: Claude Desktop, via `mcp-bridge.cjs`

## Symptom

Thunderbird Error Console:

```
Failed to start MCP server: Error: thunderbird-mcp tmp directory has group/world
permissions — refusing to write connection info
    writeConnectionInfo …/mcp_server/api.js:2169
    startPromise   …/mcp_server/api.js:8533
    start          …/mcp_server/api.js:8561
```

`%TEMP%\thunderbird-mcp\` stays empty. The client connects and completes
`initialize`, then `tools/list` fails — the bridge has no connection file to read:

```
[thunderbird-mail] [info] Server started and connected successfully
[thunderbird-mail] [info] Message from client: method="tools/list" id=1
[thunderbird-mail] [info] Message from server: id=1 error(code=-32700)
```

The extension's own options page unhelpfully shows **Running** with **Port: --**
— see the companion issue for why.

## Root cause

`api.js:2156-2176` applies POSIX permission hardening to the temp directory. The
comment states the assumption that makes it safe on Windows:

```js
// permissions is 0 on platforms that don't expose POSIX modes
// (Windows ACLs), so the chmod is a no-op there.
```

That assumption does not hold. On Windows `nsIFile.permissions` returns a
synthesized POSIX-style mode (derived from the read-only attribute), not `0`. The
consequence is a guaranteed throw:

1. `const mode = tmpDir.permissions` → non-zero, with group/world bits set
2. `(mode & 0o077) !== 0` → true, so the hardening branch is entered
3. `tmpDir.permissions = 0o700` → silently does nothing (Windows uses ACLs)
4. the re-check sees the same bits → `throw`

Steps 1 and 3 are confirmed empirically by the stack trace above: the error at
line 2169 can only be reached if `permissions` was truthy with group/world bits
*and* the assignment failed to clear them.

Worth noting the irony: this hardening defends against a hostile shared `/tmp`.
[...]

Issue #177, opened 2026-07-25
MCP server never starts on Windows: "tmp directory has group/world permissions"


raphaelmoix-creator:
**Version:** 0.7.4 (also affects every version carrying the POSIX hardening block)
**Platform:** Windows 11 Home 26200, Thunderbird 153.0, native install (not Snap/Flatpak)

## Summary

On Windows the extension's HTTP server fails to start on every run where
`%LOCALAPPDATA%\Temp\thunderbird-mcp` already exists — which is every run after
the first. `connection.json` is never written, so the bridge cannot discover the
port and the client sees no tools at all.

The POSIX hardening block in `writeConnectionInfo` is unsatisfiable on Windows.

## Error

Thunderbird Error Console:

```
WebExtensions: Failed to start MCP server: Error: thunderbird-mcp tmp directory
has group/world permissions — refusing to write connection info
  writeConnectionInfo  .../mcp_server/api.js:2169
  startPromise         .../mcp_server/api.js:8533
  start                .../mcp_server/api.js:8561
```

And from the bridge side:

```
Bridge error: Connection discovery failed. Tried: native tmp
(C:\Users\<user>\AppData\Local\Temp\thunderbird-mcp\connection.json): file not found.
Is Thunderbird running with the MCP extension?
```

## Root cause

`extension/mcp_server/api.js` (`writeConnectionInfo`, around line 2149):

```js
if (!tmpDir.exists()) {
  tmpDir.create(Ci.nsIFile.DIRECTORY_TYPE, 0o700);   // no check on this path
} else if (tmpDir.isSymlink()) {
  ...
} else {
  // permissions is 0 on platforms that don't expose POSIX modes
  // (Windows ACLs), so the chmod is a no-op there.
  const mode = tmpDir.permissions;
  if (mode && (mode & 0o077) !== 0) {
    try { tmpDir.permissions = 0o700; } catch {}
    if ((tmpDir.permissions & 0o077) !== 0) {
      throw new Error("...has group/world permissions — refusing to write connection info");
    }
  }
}
```

The comment's assumption is wrong. On Windows:

1. `nsIFile.permissions` does **not** return 0 — Gecko emulates POSIX modes and
   reports `0o777` for directories. So `mode` is truthy and `mode & 0o077` is
   non-zero.
2. The `permissions` **setter** only toggles the read-only attribute; real access
   control is by ACL. Assigning `0o700` cannot clear the emulated group/world
   bits.
[...]

Issue #171, opened 2026-07-19
[Windows] Server never writes connection.json — "tmp directory has group/world permissions" (Port: --)


ludovicricher-tech:
Title: [Windows] Server never writes connection.json — "tmp directory has group/world permissions" (Port: --)

Environment:
- Windows 11, native Thunderbird install (C:\Program Files\Mozilla Thunderbird)
- Extension build v0.7.4 (+dirty 2026-07-14)

Symptom:
- The MCP bridge reports it cannot find connection.json in %TEMP%\thunderbird-mcp.
- In Options > Server Status, status is "Running" but Port shows "--".
- Clicking "Save" surfaces: "Error: thunderbird-mcp tmp directory has group/world permissions — refusing to write connection info".

Root cause (writeConnectionInfo):
- When the tmp directory already exists, the POSIX-hardening branch runs.
- On Windows, nsIFile.permissions returns a synthesized non-zero mode (e.g. 0o777), not 0.
  So `mode && (mode & 0o077) !== 0` is true.
- Setting `tmpDir.permissions = 0o700` is a no-op against Windows ACLs, so the re-check
  still sees 0o077 bits and throws. The inline comment assumes `.permissions` is 0 on
  Windows, which is not true on this machine.

Consequence:
- It only works when the folder is absent at startup (the create() path uses mode 0o700
  and skips the check). Because the folder persists, every later Thunderbird start fails.
- Current workaround: delete %TEMP%\thunderbird-mcp before each launch.

Suggested fix:
- Skip the group/world permission check on Windows (guard by platform), or exempt
  platforms where chmod is a no-op.

TheKanishkaMohan:
Unfortunately, the suggested workaround (delete %TEMP%\thunderbird-mcp) did not work for me.

electman-AVR:
Confirming on Windows 10 Pro, v0.7.4 — same root cause, second platform.

Adding the bridge-side signature, which isn't in the original report. The
failure surfaces as a JSON-RPC parse error on tools/list roughly five
seconds after initialize succeeds:

  [thunderbird] [info] Message from client: method="tools/list" id=1 params
  [thunderbird] [info] Message from server: id=1 error(code=-32700)

The five seconds is CONNECTION_MAX_RETRIES (5) x CONNECTION_RETRY_DELAY_MS
(1000) in mcp-bridge.cjs — the bridge retrying a connection file that was
never written, then emitting a non-JSON error. Worth noting because -32700
[...]

PR #183, opened 2026-07-29
fix(windows): skip POSIX tmp dir hardening where modes aren't POSIX

### LINKED ISSUE #182: Windows: writeConnectionInfo always throws once the tmp directory outlives connection.json
## Summary

On Windows, `writeConnectionInfo()` throws unconditionally whenever the `<TmpD>/thunderbird-mcp` directory already exists:

```
    writeConnectionInfo .../mcp_server/api.js:2169
    startPromise .../mcp_server/api.js:8533
    start .../mcp_server/api.js:8561
```

Because the directory normally outlives `connection.json`, this makes the MCP server **permanently unusable** after the first time the connection file is removed without its parent directory — the server can never write the file again, and no restart recovers it.

## Root cause

The POSIX hardening block added in 7013f9f (`fix(security): tighten default policies and input validation`) rests on this assumption:

```js
// permissions is 0 on platforms that don't expose POSIX modes
// (Windows ACLs), so the chmod is a no-op there.
const mode = tmpDir.permissions;
if (mode && (mode & 0o077) !== 0) {
  try { tmpDir.permissions = 0o700; } catch { /* best-effort */ }
  if ((tmpDir.permissions & 0o077) !== 0) {
    throw new Error("thunderbird-mcp tmp directory has group/world permissions — refusing to write connection info");
  }
}
```

`nsIFile.permissions` is **not** 0 on Windows. `nsLocalFileWin::GetPermissions` returns a mode *synthesized* from the read-only attribute — typically `0o666`, or `0o777` for a directory — so `mode & 0o077` is always non-zero. `SetPermissions` on Windows only toggles the read-only attribute and cannot express POSIX group/world bits, so the `chmod` to `0o700` never sticks and the re-check fails too. The result is an unconditional throw on every Windows install.

This is unrelated to ACLs — no ACL configuration can make the synthesized mode satisfy the check.

## Why it looks intermittent

The check lives in the `else` branch, so a **fresh** directory skips it entirely:

```js
if (!tmpDir.exists()) {
  tmpDir.create(Ci.nsIFile.DIRECTORY_TYPE, 0o700);   // ← check skipped, write succeeds
} else if (tmpDir.isSymlink()) {
  ...
} else {
  // ← pre-existing directory: check runs, always throws on Windows
}
```

So a first run on a clean profile works, and the breakage only appears later. Two distinct failure modes follow, which is what makes this hard to spot:

- **In the 30s refresh timer** (`startConnectionInfoRefresh`) the throw is caught and only `console.warn`'d. The server keeps listening on its port with **no connection file on disk**, so every bridge request fails discovery while the server looks perfectly healthy. I had this state for two days — `netstat` showed the listener, `tools/list` over HTTP returned 200 with a token, but the bridge could not find the file.
- **At startup** the throw is fatal: the `catch` at `startPromise` calls `server.stop()`, so nothing binds at all and the extension is silently dead.

## Reproduction

On Windows with the tmp directory already present:

```powershell
# 1. Start Betterbird/Thunderbird once so the extension creates the directory + file
# 2. Delete only the file, leaving the directory:
Remove-Item "$env:TEMP\thunderbird-mcp\connection.json"
# 3. Toggle the add-on off/on (or restart) -> server fails to start, error above
```

Deleting the whole directory (`Remove-Item "$env:TEMP\thunderbird-mcp" -Recurse`) restores service, confirming the `!exists()` path is the only one that works on Windows.

## Environment

- thunderbird-mcp 0.7.4
- Betterbird 153.0 (`WINNT`), Windows 10 22H2
- Bridge: `mcp-bridge.cjs`, Node v22.23.1

## Suggested fix

Skip the POSIX hardening where `nsIFile.permissions` does not carry real POSIX mode bits. The shared-`/tmp` race it defends against does not apply on Windows, where the temp directory is already per-user (`%LOCALAPPDATA%\Temp`) and ACL-protected. POSIX platforms keep the existing behaviour unchanged.

Happy to send a PR — I have the fix and regression tests ready locally.

AlexTu2:
Fixes #182.

On Windows, `writeConnectionInfo()` throws every time the `<TmpD>/thunderbird-mcp` directory already exists, which makes the MCP server permanently unusable once `connection.json` is removed without its parent directory.

## The bug

The POSIX hardening added in 7013f9f assumes `nsIFile.permissions` is `0` where POSIX modes don't exist:

```js
// permissions is 0 on platforms that don't expose POSIX modes
// (Windows ACLs), so the chmod is a no-op there.
```

That isn't true on Windows. `nsLocalFileWin::GetPermissions` synthesizes a mode from the read-only attribute — typically `0o666`, or `0o777` for a directory — so `mode & 0o077` is always non-zero. `SetPermissions` can only toggle the read-only attribute and cannot express POSIX group/world bits, so the `chmod` to `0o700` never sticks and the re-check throws as well. No ACL configuration can satisfy the check.

## Why it presents as intermittent

The check is in the `else` branch, so a **fresh** directory skips it and works fine. The breakage appears only after `connection.json` is deleted while its parent directory survives — and then it is permanent, in two different disguises:

- **Refresh timer:** the throw is caught and only `console.warn`'d, so the server keeps listening on its port with no connection file on disk. The server looks healthy (port bound, authenticated `tools/list` returns 200) while every bridge discovery fails. I sat in this state for two days before finding the error console.
- **Startup:** the throw is fatal — `startPromise`'s catch calls `server.stop()`, so nothing binds and the extension is silently dead.

## The fix

Extract the hardening into a top-level `hardenTmpDirPermissions()` helper and skip it where `nsIFile.permissions` carries no real POSIX mode. Windows temp is already per-user (`%LOCALAPPDATA%\Temp`) and ACL-protected, so the shared-`/tmp` race this guards against doesn't apply there.

**POSIX behaviour is unchanged** — the guard clause is exactly the inverse of the original condition (`!(mode && (mode & 0o077) !== 0)` ≡ `!mode || (mode & 0o077) === 0`), and the chmod-then-verify-then-throw sequence is preserved as-is.

The helper is placed in a `BEGIN/END TMP DIR HARDENING HELPERS` marked block so it's loadable from tests via `vm`, matching the existing `CONNECTION INFO REFRESH HELPERS` convention.

## Tests

New `test/tmp-dir-hardening.test.cjs` (10 cases), including a regression test asserting that Windows neither throws nor attempts a chmod when the synthesized mode looks permissive:

- `tmpDirModeIsPosix` — `WINNT` is non-POSIX; Linux/Darwin/FreeBSD/OpenBSD are POSIX
- Windows: skips hardening, no chmod attempted (**regression**)
- POSIX `0o700`: left alone, no chmod
- POSIX mode `0`: treated as an unsupported accessor, not a permissive dir
- POSIX `0o755`: chmod'd to `0o700`
- POSIX chmod that doesn't stick: still throws

Full suite: **527 tests, 509 passed, 0 failed, 18 skipped**. `npm run lint`: 0 errors (13 pre-existing warnings, none in touched code).

## Verification

Reproduced and fixed on Betterbird 153.0 / Windows 10 22H2 with 0.7.4. Before: server failed to start with the `group/world permissions` error whenever the directory pre-existed. After deleting the directory to confirm the `!exists()` path was the only working one, the server bound `127.0.0.1:8766` and returned 40 tools over an authenticated `tools/list`.

I have not rebuilt `extension.xpi` — the repo appears to regenerate that in separate `build:` commits, so I left it to you.


justerlex:
Same failure here, on Thunderbird 153.0.1 (Windows 11) with extension 0.7.4. One detail I haven't seen spelled out in the issues, and I think it explains why there are so many of them.

The bug survives everything a user tries first: toggling the extension off and on, reinstalling the XPI, restarting Thunderbird. Deleting `%TEMP%\thunderbird-mcp` does fix it, but only for one session, because the next start then takes the `create` branch and never reaches the permission check. Restart after that and the directory is back, so it throws again. It cost me two days of blaming my own machine before I opened the error console. Six open issues describe the same failure in different words (#167, #171, #177, #178, #181, #182), which fits that pattern.

With this PR's change applied locally, before and after on my box:

- before: no `connection.json`, no listening socket owned by `thunderbird.exe`, ports 8765-8774 all free, and the bridge reporting `Connection discovery failed ... file not found` while Thunderbird sat there running with the extension enabled.
- after: `connection.json` written on startup, Thunderbird listening on `127.0.0.1:8765`, and a direct `tools/list` against the HTTP API returning all 40 tools.

I had written the same fix locally before I found this PR: same `os !== "WINNT"` gate, same marker-block idiom, tests in the same style. So call it independent confirmation rather than a competing patch. `hardenTmpDirPermissions` returning `{hardened, reason}` is a better shape than the two loose predicates I ended up with, so I am dropping mine. Either this or #170 gets Windows working again, and I do not much mind which one lands.

PR #192, opened 2026-08-09
fix(filters): correct nsMsgSearchAttrib/nsMsgFilterAction values and typed condition values

ncrosty58:
## Summary

The filter constant maps in `api.js` assume Thunderbird's two enums are dense runs starting at 0. Neither is, so most names resolve to the wrong constant. Separately, `buildTerms` writes every condition through `nsIMsgSearchValue.str`, which makes date and numeric conditions impossible to create.

I hit this trying to build a date-based archive filter: `createFilter` threw `NS_ERROR_ILLEGAL_VALUE [nsIMsgSearchValue.str]` at `buildTerms`, and a filter created with `stopExecution` was written to `msgFilterRules.dat` as `action="Delete from Pop3 server"`.

## What was wrong

**`ATTRIB_MAP`** — correct through `allAddresses` (9), wrong from there on:

| name | resolved to | should be |
|---|---|---|
| `ageInDays` (10) | `Location` | 12 |
| `size` (11) | `MessageKey` | 14 |
| `tag` (12) | `AgeInDays` | 16 |
| `hasAttachment` (13) | `FolderInfo` | 44 |
| `junkStatus` (14) | `Size` | 45 |
| `junkPercent` (15) | `AnyText` | 46 |
| `otherHeader` (16) | `Keywords` | 52 |

**`ACTION_MAP`** — correct only for `moveToFolder`, shifted by one from `changePriority` on:

| name | resolved to | should be |
|---|---|---|
| `copyToFolder` (0x02) | `ChangePriority` | 16 |
| `changePriority` (0x03) | `Delete` | 2 |
| `delete` (0x04) | `MarkRead` | 3 |
| `markRead` (0x05) | `KillThread` | 4 |
| `markFlagged` (0x08) | `Label` | 7 |
| **`stopExecution` (0x0C)** | **`DeleteFromPop3Server`** | 11 |
| `junkScore` (0x0F) | `FetchBodyFromPop3Server` | 14 |
| `markUnread` (0x14) / `custom` (0x15) | undefined values | 19 / -1 |

`stopExecution` is the one worth flagging: asking a filter to stop processing instead installed a delete-from-server action. That is inert on IMAP but destructive on POP3.

`label` and `deleteBody` are dropped, as neither names a current action. `custom` is dropped because it needs a `customId` that this API has no way to supply.

**`buildTerms`** wrote `value.str` unconditionally. `nsIMsgSearchValue` is a tagged union keyed on `attrib`, so any date or numeric attribute threw `NS_ERROR_ILLEGAL_VALUE` — no date, age, size, priority, status or junk condition could be created at all.

**Date handling.** Date-only strings are now read as a local calendar day. `Date.parse()` treats them as UTC midnight, which Thunderbird renders back in local time, so `2024-01-01` was stored as `31-Dec-2023` anywhere west of UTC. Strings carrying explicit timezone information still go through `Date.parse()` unchanged.

The reader side (`serializeFilter`) had the same hardcoded assumptions and is updated to match, so filters round-trip.

## Verification

Against Thunderbird 140.13.0esr, creating filters through the MCP and reading back `msgFilterRules.dat`:

```
action="Stop execution"                              # was "Delete from Pop3 server"
action="Mark read"                                   # was "Kill thread"
condition="AND (age in days,is greater than,952)"    # previously threw
condition="AND (date,is before,01-Jan-2024)"         # previously threw; also was 31-Dec-2023 pre-timezone-fix
```

## Tests

`test/filter-enum-values.test.cjs` parses `ATTRIB_MAP`, `ACTION_MAP` and `NUMERIC_VALUE_FIELDS` out of `api.js` rather than restating them, so the constants cannot drift from the IDL without failing. Includes explicit regression guards for the `stopExecution`/`DeleteFromPop3Server` collision and the UTC date shift.

`npm test`: 530 tests, 512 pass, 1 fail — the pre-existing `auth.test.cjs:252 "fails with wrong token"`, unrelated and failing on `main` too. `npm run lint`: 0 errors.

## Note

I did not rebuild `dist/thunderbird-mcp.xpi` in this PR, since a binary diff is not reviewable — it will need regenerating via `scripts/build.sh` before release.

Values sourced from `mailnews/search/public/nsMsgSearchCore.idl` and `nsMsgFilterCore.idl`.


ncrosty58:
Closing as a duplicate of #175, which was opened first and covers the same enum and tagged-union fixes. I've left the details there.

Two findings from this branch are worth carrying over, and I've commented on #175 with both:

- **Date conditions are written a day early west of UTC.** `Date.parse("2024-01-01")` is UTC midnight, but Thunderbird stores filter dates in local time, so it lands as `31-Dec-2023`. #175 has the same issue. Fix and reproduction are in the comment.
- **Tests pinning the constants to the IDL**, so the values can't drift silently: https://github.com/ncrosty58/thunderbird-mcp/blob/fix-filter-enum-values/test/filter-enum-values.test.cjs

The branch stays up if any of it is useful. #175's `hasAttachment` handling is more correct than what I had here, and I've adopted it with credit.

Issue #205, opened 2026-08-28
Windows: server never binds — "tmp directory has group/world permissions" (nsIFile.permissions is not 0 on Windows)


piotrekkonopka100-sys:
**Summary**
On Windows the MCP server never binds a port. The options page shows Server Status: Running but Port: --, no connection.json is written, and nothing listens on 8765-8774.

**Environment**
Windows 11 (build 26200)
Thunderbird 154.0
Extension v0.7.4 +3 (80dc749), built 2026-08-27
**Error (Error Console)**
```
WebExtensions: Failed to start MCP server: Error: thunderbird-mcp tmp directory
has group/world permissions — refusing to write connection info
    writeConnectionInfo .../mcp_server/api.js:2169
    startPromise .../mcp_server/api.js:8544
    start .../mcp_server/api.js:8572
```
**Root cause**
writeConnectionInfo() applies POSIX hardening to <TmpD>/thunderbird-mcp. The code comment states:

> permissions is 0 on platforms that don't expose POSIX modes (Windows ACLs), so the chmod is a no-op there.

That assumption does not hold. On Windows nsIFile.permissions returns a synthesised POSIX-like mode (directories typically report 0777), not 0. So:

mode && (mode & 0o077) !== 0 evaluates to true
tmpDir.permissions = 0o700 is silently ineffective (NTFS ACLs)
the re-check still sees group/world bits and throws
Startup aborts before the port-binding loop, which is why the status flag reads "Running" while Port stays empty.

Ports 8765-8774 were verified to be free, so this is not a port conflict.

**Why it may not reproduce immediately**
The check lives in the else branch, so it runs only when the directory already exists. A first run on a clean machine creates it with 0o700 and skips the check entirely; failures appear once the directory persists.

**Workaround**
Delete %TEMP%\thunderbird-mcp and restart the extension. Works until the directory exists again, so it has to be repeated before every Thunderbird start.

**Suggested fix**
Skip the POSIX hardening on Windows, for example:

```
} else if (Services.appinfo.OS !== "WINNT") {
  // POSIX hardening: ...
}
```
The hardening protects against a shared /tmp. On Windows each user has a private %TEMP%, so the check guards against a threat that does not exist there while blocking every Windows user.

**Minor, separate issue**
The options page reports Server Status: Running even though startup threw and no socket was opened. Reporting the actual bind state would have made this much faster to diagnose.

Issue #203, opened 2026-08-25
Windows: MCP server starts only once, POSIX tmp dir check rejects every later start


andresolbach:
## Summary

On Windows the MCP server starts **exactly once**, on the run that first creates `<TmpD>\thunderbird-mcp`, and fails on every start after that. No port is bound, no `connection.json` is written, and the bridge reports:

```
Bridge error: Connection discovery failed. Tried: native tmp
(C:\Users\<user>\AppData\Local\Temp\thunderbird-mcp\connection.json): file not found.
Is Thunderbird running with the MCP extension?
```

The extension itself looks perfectly healthy in the Add-ons Manager, which makes this hard to spot. It is enabled, not blocklisted, and `addonStartup.json.lz4` lists it as enabled. The failure is silent unless you happen to have the Browser Console open, where it shows up as:

```
Failed to start MCP server:
Error: thunderbird-mcp tmp directory has group/world permissions — refusing to write connection info
  writeConnectionInfo@jar:file:///...thunderbird-mcp@tkasperczyk.dev.xpi!/mcp_server/api.js:2169:29
  getAPI/start/startPromise<@...api.js:8533:30
```

## Root cause

The POSIX hardening in `writeConnectionInfo()` (extension/mcp_server/api.js):

```js
// permissions is 0 on platforms that don't expose POSIX modes
// (Windows ACLs), so the chmod is a no-op there.
const mode = tmpDir.permissions;
if (mode && (mode & 0o077) !== 0) {
  try { tmpDir.permissions = 0o700; } catch { /* best-effort */ }
  if ((tmpDir.permissions & 0o077) !== 0) {
    throw new Error("thunderbird-mcp tmp directory has group/world permissions — ...");
  }
}
```

The comment assumes `nsIFile.permissions` is `0` on Windows. It isn't. `nsLocalFileWin::GetPermissions` synthesises a Unix-ish mode from the file attributes and returns `0o777` for any writable directory, or `0o555` when the read-only attribute is set. `SetPermissions` on Windows only toggles that read-only attribute, so assigning `0o700` cannot clear the group/world bits. They read straight back as `0o777`.

The guard can therefore never be satisfied on Windows and the throw is unconditional. It only stays invisible on the very first run, because that path takes the `create(DIRECTORY_TYPE, 0o700)` branch and skips the check entirely.

## Reproduction

Thunderbird 153.0.3 (release, Windows 11), extension v0.7.4, fresh throwaway profile:

[...]

Issue #202, opened 2026-08-24
Server fails to start on Windows when the tmp directory already exists (POSIX permission check false-positives on NTFS)


nobotaro-code66:
## Environment

- Windows 11 Pro (10.0.26200)
- Thunderbird 154.0 (ja), native install
- thunderbird-mcp v0.7.4 (pre-built XPI from `dist/`)

## Symptom

On Windows, the MCP server starts only on the *first* launch after `%LOCALAPPDATA%\Temp\thunderbird-mcp` has been removed (e.g. after an OS temp cleanup). On every subsequent Thunderbird launch the server silently fails to start:

- No port in 8765–8774 is listening, `connection.json` is never written
- The bridge reports `MCP error -32000: Connection closed` to the client
- The options page misleadingly shows **Running** with port `--` (see "Related bug" below)
- Nothing is logged to the global Error Console

Running `await browser.mcpServer.start()` in the add-on debugging console returns:

```
{ success: false, error: "Error: thunderbird-mcp tmp directory has group/world permissions — refusing to write connection info" }
```

## Root cause

`writeConnectionInfo()` in `extension/mcp_server/api.js` runs a POSIX hardening check when the tmp directory already exists:

```js
const mode = tmpDir.permissions;
if (mode && (mode & 0o077) !== 0) {
  try { tmpDir.permissions = 0o700; } catch { /* best-effort */ }
  if ((tmpDir.permissions & 0o077) !== 0) {
    throw new Error("thunderbird-mcp tmp directory has group/world permissions — refusing to write connection info");
  }
}
```

The comment above this block assumes `permissions` is `0` on Windows, but in practice `nsIFile.permissions` on Windows returns *synthesized* mode bits derived from the read-only attribute (`0o666`/`0o777`-style values), never `0`. Writing `permissions = 0o700` only toggles the read-only attribute, so the re-read still contains group/world bits and the check always throws.

Because the check only runs in the "directory already exists" branch, the failure mode is intermittent in a confusing way:

- fresh dir (first run, or after temp cleanup / reboot with cleanup policy) → create path, no check → **works**
- dir left over from a previous run → check runs → **always fails on Windows**

## Suggested fix

Scope the check to POSIX platforms; on Windows the per-user `%TEMP%` is already protected by NTFS ACLs:

```js
try {
  // nsIFile.permissions on Windows synthesizes mode bits from the
  // read-only attribute (e.g. 0o666), so the POSIX check below would
[...]

PR #206, opened 2026-09-01
fix(windows): skip POSIX tmp-dir permission check that always fails

### LINKED ISSUE #167: v0.7.4 fails to start on Windows: tmp-dir permissions check can never pass (nsIFile.permissions reports 0o666 for any directory)
**Repro:** Windows 11, Thunderbird 140 ESR, install 0.7.4, Claude Code CLI and Cowork (sandbox). First TB start works (dir freshly created). Every subsequent start fails: `Failed to start MCP server: Error: thunderbird-mcp tmp directory has group/world permissions — refusing to write connection info`. The shutdown cleanup removes only `connection.json`, never the directory, so the failure is deterministic from the second start on.

**Cause:** `writeConnectionInfo()` checks `tmpDir.permissions & 0o077` on the existing dir. On Windows, Gecko's `nsLocalFileWin::GetPermissions` unconditionally reports group/world read bits and, for any directory, write bits too (`IsWritable` → `IsDirectory` → true), i.e. 0o666 — regardless of ACLs. `SetPermissions(0o700)` maps to `_wchmod`, which only toggles FILE_ATTRIBUTE_READONLY, so the re-check always fails. The in-source comment "permissions is 0 on platforms that don't expose POSIX modes (Windows ACLs)" does not hold for Gecko on Windows. Ref: mozilla-firefox/firefox `esr140`, `xpcom/io/nsLocalFileWin.cpp` (GetPermissions ~2551, SetPermissions ~2616, IsWritable ~2872). Introduced in 7013f9f (PR #125), first shipped in v0.7.4.

**Note:** `extensions.thunderbird-mcp.stableAuthToken` is not a workaround — `writeConnectionInfo` runs unconditionally on every server start.

**Suggested fix:** skip the POSIX mode check on Windows (`Services.appinfo.OS === "WINNT"`); the shared-/tmp threat model it defends against does not exist there (per-user, ACL-protected %TEMP%). Alternatively treat "mode has group/world bits AND platform exposes real POSIX modes" as the trigger.

ccrtr:
\`writeConnectionInfo()\` reads \`nsIFile.permissions\` to harden the connection-info tmp directory against a shared-\`/tmp\` race on POSIX systems, assuming (per the code's own comment) that this reads as \`0\` on platforms without POSIX modes — meaning the check should be a no-op on Windows.

In practice, Gecko synthesizes a POSIX-like mode for NTFS directories (typically \`0777\`). The check is therefore always truthy, \`chmod\` is a silent no-op against ACLs, and the re-check always throws — so the extension can never write \`connection.json\` on Windows. Confirmed via \`Services.appinfo.OS\`; fixes #167, #171, #177, #178, #181, #182, #202, #203, #205.

This skips the POSIX branch on \`Services.appinfo.OS === "WINNT"\`. The symlink check and the \`O_CREAT|O_EXCL\` exclusive create are untouched on every platform. The multi-user shared-\`/tmp\` threat this hardening defends against doesn't apply on Windows, where \`%TEMP%\` is already private per-user via NTFS ACLs — which the POSIX mode check was never able to observe anyway.

## Test plan
- [x] Built the patched XPI and loaded it on Windows 11
- [x] Extension now writes \`connection.json\`; options panel shows a real port instead of \`--\`
- [x] Bundled \`mcp-bridge.cjs\` connects and returns the full tool list
- [x] Claude Desktop connects successfully with all tools available
- [x] \`saveDraft\` with a file-path attachment succeeds and the attachment is present on the saved draft


static9:
Confirming this fixes the bug on Windows 11. Before this patch the server never wrote `connection.json` unless I deleted `%TEMP%\thunderbird-mcp` before every launch — on NTFS `nsIFile.permissions` returns a synthesized `0777`, so the `(mode & 0o077)` check in `writeConnectionInfo()` throws whenever the directory already exists. Gating that block with `Services.appinfo.OS !== "WINNT"`, as this PR does, resolves it — the server now starts cleanly against a pre-existing temp dir. Would be great to get this merged. Thanks for the work!

Issue #197, opened 2026-08-22
Server Status shows "Running" but Port is empty and connection.json is never written (Windows)


cdpthan:
**Environment**
- OS: Windows (win32, x64)
- Thunderbird MCP extension version: 0.7.4
- Build shown on the extension's own Options page: `v0.7.4 (v0.7.4) +dirty 2026-07-14 19:42:11Z`
- Install method: cloned the repo, installed `dist/thunderbird-mcp.xpi` via "Install Add-on From File"
- MCP client: Claude desktop app, local MCP server config:
```json
  {
    "command": "C:\\Program Files\\nodejs\\node.exe",
    "args": ["C:\\Users\\<user>\\thunderbird-mcp\\mcp-bridge.cjs"]
  }
```

**Bug**
The bridge (`mcp-bridge.cjs`) fails on every attempt with:

Bridge error: Connection discovery failed. Tried: native tmp
(C:\Users<user>\AppData\Local\Temp\thunderbird-mcp\connection.json): file not found.
Is Thunderbird running with the MCP extension? The extension must be
started first to create the connection file.

On the extension's own **Options → Server Status** page it shows:
- Status: 🟢 Running
- Port: `--` (empty, never populated)
- Connection file: `C:\Users\<user>\AppData\Local\Temp\thunderbird-mcp\connection.json` (correct expected path)
- Build: `v0.7.4 (v0.7.4) +dirty 2026-07-14 19:42:11Z`

The temp folder `C:\Users\<user>\AppData\Local\Temp\thunderbird-mcp\` exists but is always empty — `connection.json` is never written, despite the status page claiming the server is running.

**Steps already tried (no change in behavior)**
- Full Thunderbird restart (verified no lingering `thunderbird.exe` process in Task Manager) — multiple times
- Full Claude desktop app restart
- Complete removal of the extension + reinstall from the `.xpi` + restart (per the README's "dev-only extension reload" instructions)
- Checked Thunderbird's general Error Console — empty, no errors logged
- Checked the extension's dedicated console via `about:debugging` → multiprocess browser toolbox → Inspect → Console — no log output at all, even when toggling the extension off/on with the console open
- Checked `netstat -ano | findstr "876"` — confirms nothing is listening on ports 8765–8774, so it's not a port conflict
- Checked Windows Firewall "Allow an app" list for `node.exe` and `thunderbird.exe` — enabled for both Private/Public
[...]

PR #209, opened 2026-09-11
fix(server): skip POSIX tmp-dir permission check on Windows

x4121t:
## Problem

On Windows the MCP server never binds once `%TEMP%\thunderbird-mcp` exists. The options page shows "Running" with an empty port, `connection.json` is never written, and the Error Console reports:

```
```

Reported in #178, #181, #182, #197, #202, #203, #205.

## Root cause

`writeConnectionInfo()` applies POSIX hardening to the connection directory. The comment assumes `nsIFile.permissions` returns 0 on Windows, but it returns a synthesised mode (directories report `0o777`), and assigning `0o700` is a no-op against NTFS ACLs. So the re-check always still sees group/world bits and throws. The check lives in the `else` branch, which is why a first run on a clean machine works and every later start fails (#182, #203).

## Fix

Skip the POSIX branch when `Services.appinfo.OS === "WINNT"`. `%TEMP%` is per-user and ACL-protected on Windows, so the shared-`/tmp` race the check guards against does not apply there. Behaviour on Linux/macOS is unchanged. The symlink check still runs on all platforms.

## Test

Added `test/write-connection-info-windows.test.cjs`, which extracts `writeConnectionInfo` via `BEGIN/END CONNECTION INFO WRITER` markers (same pattern as the other tests) and covers:

- Windows, directory reports `0o777`, chmod ineffective → writes the file, no chmod attempted
- Linux, `0o777`, chmod ineffective → still throws
- Linux, `0o755`, chmod works → repairs to `0o700` and writes

`node --test test/*.cjs`: 528 pass, 0 fail. `eslint` reports no new warnings.

## Verified on

Windows 10 Pro 19045, Thunderbird 153.2.0, v0.7.5 + this patch built with `scripts/build-xpi.cjs`:

- Fresh start: `connection.json` written, server listening on 127.0.0.1:8765, `tools/list` returns 40 tools, `listAccounts` works through `mcp-bridge.cjs`.
- Existing-directory path (the failing case): deleted `connection.json` while Thunderbird kept running; the refresh timer recreated it through the patched branch instead of throwing.

The `dist/` XPI is intentionally not included in this PR.


sacrefizz:
Thanks for this fix. I compared the eight PRs for this bug (#170, #183, #189, #191, #199, #204, #206, #209), and this one looks like the easiest to review: the production change is a single `else if` guard, POSIX behaviour is unchanged, and the new test exercises the real `writeConnectionInfo` on both the Windows and Linux paths.

I'm running this exact change locally (Windows 11 26200, Thunderbird 156.0.1, 0.7.5): with an existing `%TEMP%\thunderbird-mcp` folder, `connection.json` is rewritten and the bridge connects (details in #212).

All 529 tests pass on this branch (`node --test`, run offline).

Two small notes: the issue numbers in the code comment would fit better in the PR description, and #170 and #206 also modify `dist/thunderbird-mcp.xpi`, which can't be reviewed from source.

Credit also goes to the earlier work on this bug: #170 was first to propose this guard, and #183 has the most complete analysis (including the refresh-timer path).

<sub>Drafted with help from an AI assistant (Claude), reviewed and owned by @sacrefizz.</sub>

PR #218, opened 2026-09-22
feat(compose): report the drafts folder saveDraft wrote to

ajoe:
## Problem

`saveDraft` resolves with `{success, message}` only. A caller that wants to look the draft up afterwards — to tag it, to show it to the user — has to guess which folder it went to. It passed `from`, but the mapping from an identity to its drafts folder is Thunderbird's business, not the caller's.

## Change

Report it as `folderPath` on success, read off the identity.

The property name has moved across Thunderbird versions, so the known ones (`draftsFolderURI`, `draftFolder`, `draftsFolder`, `fccFolder`) are probed in turn, and either a URI string or a folder object is accepted. A version that exposes none of them simply returns the previous shape.

Best effort throughout: a missing or throwing property never turns a successful save into an error. Existing callers see the same result plus one new field.

## Tests

`test/savedraft-folder-path.test.cjs`, VM-extracted from production like the existing suites. Covers the URI-string and folder-object shapes, a property that throws instead of existing, a value that is not a folder URI, and an identity that names no drafts folder at all.

`node --test test/*.cjs` — 531 tests, 0 failures.

## Note

Independent of #217, which touches the same tool but different lines.

The branch is still named `feat/savedraft-returns-message-id` from an earlier, wider version of this PR — see the comment below. GitHub does not allow renaming a PR's head branch, and closing this in favour of a cosmetically better name seemed like more noise than it is worth.


ajoe:
I have force-pushed a substantially narrower version of this PR, and I want to be upfront about why.

The original version also returned the saved draft's `messageId`. It captured the message key from `nsIMsgCopyServiceListener.setMessageKey()` and the folder from `nsIMsgSendListener.onGetDraftFolderURI()` — both of which are empty stubs in `sendMessageDirectly`, and the code comment right above them says completion for `SaveAsDraft` is signalled via `onStopCopy`.

Testing it against a real Thunderbird showed that is not what happens, at least not on every version: **for the `SaveAsDraft` path, `onStopCopy` never fired, and neither did `setMessageKey` or `onGetDraftFolderURI`.** The feature was inert. I only caught it because a caller downstream reported the draft could not be found — the unit tests passed, because they stubbed the listener I had assumed into existence.

So this PR is now limited to what I can actually demonstrate: the drafts folder, read off the identity rather than from a listener callback.

If the `messageId` is worth having, the promising route looks like setting `composeFields.messageId` before the save rather than reading it back afterwards. I have not proposed that here because Thunderbird appears to assign drafts a fresh Message-ID on every save ([bug 1279869](https://bugzilla.mozilla.org/show_bug.cgi?id=1279869)), so whether a caller-set value survives needs to be established first — and I did not want to repeat the mistake of shipping an assumption.

Happy to close this instead if the narrowed version is not worth the diff.

